PR #1, opened 2022-03-29
add configuration for `wp-env`

fabiankaegy:
This PR adds the most barebones configuration for `wp-env` to just have the default setup with the plugin activated.


luisherranz:
Thanks, Fabian!

Out of curiosity, how did you find this repository? Are you interested in block frontend hydration? 🙂

fabiankaegy:
@luisherranz I've been following you on GitHub for quite a while now and saw that you've created the repository in the "Following" tab of the GitHub homepage =D 

But yeah I'm very interested in the topic. In the past, I've mostly been working with either using the `render` method from react dom to hydrate the frontend like here: https://github.com/fabian-kaegy-org/finn/blob/master/src/scripts/functionality/wp-block-gallery-lightbox.js
or use vanilla JS in both the editor and the frontend like so: https://github.com/10up/maps-block-apple/blob/develop/src/frontend.js

So I am very curious about any exploration into how core could solve that issue in the future. 

I don't have a ton of familiarity with custom elements so I am quite interested in exploring your solution here. 


The main issue that I've not found a great solution for is how you can create a shared component that renders as a `contenteditable` `RichText` element in the editor and as an actual non-editable version on the frontend. Because in the end, the main issue in my head is having to write all the logic and markup twice. But in order to create a nice visual editorial experience that is what you have to do currently. 

Ideally this here:
https://github.com/luisherranz/block-hydration-experiments/blob/0bad4534dbe386908ed43ed8035faec90d8a26ce/src/frontend/title.js#L1

could also be used in the editor to replace the `TextControl` here:
https://github.com/luisherranz/block-hydration-experiments/blob/0bad4534dbe386908ed43ed8035faec90d8a26ce/src/edit.js#L9-L13

with a `RichText` component that shares all the same attributes.

luisherranz:
> I've been following you on GitHub for quite a while now and saw that you've created the repository in the "Following" tab of the GitHub homepage

Nice. I didn't know about that thing! 😄 

> The main issue that I've not found a great solution for is how you can create a shared component that renders as a contenteditable RichText element in the editor and as an actual non-editable version on the frontend

Yes, that's something I'd like to investigate as well. Let's continue the conversation in the other PR.

PR #3, opened 2022-04-01
Add wrapperless blocks, hook shims and a block support

luisherranz:
Continuing the tests here, I've added support for:

- Wrapperless hydration: now, the `Block` component can control the wrapper `div` as well.
- Hook shims: I've added support for `useState` and `useEffect` in `Block` using @fabiankaegy's trick. This will need to replace replaced with something natively supported by Gutenberg. I'd like to take a look at the [custom serializer](https://github.com/WordPress/gutenberg/blob/trunk/packages/element/src/serialize.js) with @gziolo.
- Block supports (only in the wrapper node, no skip-serialization): I'm serializing the block props (which include the `has-text-color`... class names) because I thought React would remove them, but it doesn't seem to do so. I need to investigate more.

The block now hides/shows the children to test that it can destroy and recreate them while preserving hydration correctly.

### DIFF
--- a/src/framework/save.js
+++ b/src/framework/save.js
@@ -1,18 +1,21 @@
 import { name } from "../block.json";
 import { useBlockProps, InnerBlocks } from "@wordpress/block-editor";
 
-const save = (Comp) => ({ attributes }) => (
-  <div
-    is={name.replace("/", "_")}
-    {...useBlockProps.save()}
-    data-gutenberg-attributes={JSON.stringify(attributes)}
-  >
-    <Comp attributes={attributes}>
-      <gutenberg-inner-blocks>
-        <InnerBlocks.Content />
-      </gutenberg-inner-blocks>
-    </Comp>
-  </div>
-);
+const save = (Comp) => ({ attributes }) => {
+  const blockProps = useBlockProps.save();
+  return (
+    <gutenberg-interactive-block
+      data-gutenberg-block-type={name}
+      data-gutenberg-attributes={JSON.stringify(attributes)}
+      data-gutenberg-block-props={JSON.stringify(blockProps)}
+    >
+      <Comp blockProps={blockProps} attributes={attributes}>
+        <gutenberg-inner-blocks>
+          <InnerBlocks.Content />
+        </gutenberg-inner-blocks>
+      </Comp>
+    </gutenberg-interactive-block>
+  );
+};
 
 export default save;
--- a/src/framework/wordpress-element.js
+++ b/src/framework/wordpress-element.js
@@ -1,0 +1,14 @@
+import {
+  useState as useReactState,
+  useEffect as useReactEffect,
+} from "@wordpress/element";
+export { hydrate } from "react-dom";
+
+// Dirty dirty trick
+export const isView = !window.wp.blockEditor;
+
+const noop = () => {};
+
+export const useState = (init) => (isView ? useReactState(init) : [init, noop]);
+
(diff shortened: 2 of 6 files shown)


src/framework/wordpress-element.js
```diff
@@ -0,0 +1,14 @@
...
+  useEffect as useReactEffect,
+} from "@wordpress/element";
+export { hydrate } from "react-dom";
+
+// Dirty dirty trick
+export const isView = !window.wp.blockEditor;
+
+const noop = () => {};
+
+export const useState = (init) => (isView ? useReactState(init) : [init, noop]);
```
fabiankaegy, line 12: I'm curious whether your thinking here is to leave these as `noop` functions forever or whether it is using a tree shaken smaller version of react instead of the WordPress global? 🤔

luisherranz: The noop functions are for `Save` and in the Editor it doesn't make much sense to do tree-shaking (at least is not an urgent priority). In the frontend we need the real hooks.

By the way, I'd love to not use globals in the frontend. They are not needed for backwards compatibility so we can switch to regular imports 🙂

src/frontend/block.js
```diff
@@ -1,18 +1,18 @@
...
-const Block = ({ attributes, children, isView }) => {
-  const [show, setShow] = isView ? useState(true) : [true];
-  const [counter, setCounter] = isView ? useState(0) : [0];
+const Block = ({ blockProps, attributes, children }) => {
+  const [show, setShow] = useState(true);
+  const [counter, setCounter] = useState(0);
 
   return (
-    <>
+    <div {...blockProps}>
```
luisherranz, line 10: Funny thing, if we remove `blockProps`, React doesn't remove the classes present in the HTML on hydration.

I want to investigate this more.

src/framework/save.js
```diff
@@ -1,18 +1,21 @@
...
-  >
-    <Comp attributes={attributes}>
-      <gutenberg-inner-blocks>
-        <InnerBlocks.Content />
-      </gutenberg-inner-blocks>
-    </Comp>
-  </div>
-);
+const save = (Comp) => ({ attributes }) => {
+  const blockProps = useBlockProps.save();
```
fabiankaegy, line 5: I think we somehow need to expose the `useBlockProps.save` funciton call to the developer in the `frontent/block.js` file because that is the only way to add additional attributes like the className to it.

Mayve we can make it completely hidden though. What if the developer would be able to import a `Block` component that works like this:

```js
function BlockSave(props) {
    const { attributes } = props;

    return (
        <Block className="custom-class">
            <Title message={attributes.message} />
        </Block>
    );
}
```

which would call `useBlockProps.save` under the hood and spread all the props into the `useBlockProps.save` callback

luisherranz: Yes, that's definitely an implementation detail that should be hidden because it doesn't add any value to the developer. Let's see if we can get rid of it 🙂

luisherranz: I've added a [Readme](https://github.com/luisherranz/block-hydration-experiments/blob/main/Readme.md). Feel free to add more items to the list.

luisherranz: I've added a [Readme](https://github.com/luisherranz/block-hydration-experiments/blob/main/Readme.md). Feel free to add more items to the list.

luisherranz:
I'm going to merge this one to continue opening other PRs on top of it.

PR #4, opened 2022-04-04
Refactor folder to make structure more obvious

luisherranz:
Just a small refactoring of the code to make it more clear and an updated Readme.


luisherranz:
Merging this, as it's only a refactoring.

PR #5, opened 2022-04-06
Use children instead of InnerBlocks

luisherranz:
I want to open a PR to explore the difference between `children` and `<InnerBlocks>`. With this PR, all contexts (Edit and Save/Frontend) here use `children` instead of `<InnerBlocks>`.

There are some [configuration settings](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/nested-blocks-inner-blocks/) that can be abstracted in the `block.json`. I wonder how many of them need to be calculated dynamically and if passing them as props to an `<InnerBlocks>` component is the correct approach.


luisherranz:
I'm going to close this for now. We can revisit later.

PR #7, opened 2022-04-08
Support BlockContext between different blocks

luisherranz:
The idea is to try using a custom event that is captured by a parent block that populates the context.

Inspired by [preact-custom-element](https://github.com/preactjs/preact-custom-element/blob/master/src/index.js#L71-L77).


luisherranz:
I've added a small explanation of how this could work in the last summary video: https://github.com/luisherranz/block-hydration-experiments/issues/6#issuecomment-1129029827 (minute 6:30)

michalczaplinski:
Sorry, for the triple force-push - I've changed the base of this PR (to include the fix in #19) and messed up the local branch in the process. All fixed now 🙂 .

luisherranz:
No problem. I've merged `main` to keep it in sync with the latest changes (mainly formatting) 👍

luisherranz:
I'm now the one who has messed something up 😆

luisherranz:
Rebasing is not working either. Main commits are still appearing 😕

luisherranz:
Ohhhh, ok. I got it. It's because you changed the base. I'll change it back to `main`.

luisherranz:
Great. Everything back to normal 🙂

michalczaplinski:
Yes! I don't have it in a workable state yet, but I ll share an update later today 🙂

michalczaplinski:
Below's the current status update in video form.

The code is "working" but I didn't make much effort to clean it up yet. 

There is also an issue that I have mentioned in the video, which I ll document later on. When including the parent block includes **only** the child block, it does not render that child block. I *think* this might be an issue with rendering inner blocks that was introduced in #8. 

<a href="https://www.loom.com/share/a0f0d917491e471990d36d62da6cbb1e">
    
  </a>

src/gutenberg-packages/frontend.js
```diff
@@ -1,17 +1,37 @@
...
 
-const blockTypes = new Map();
+// We assign `blockTypes` to window to make sure it's a global singleton.
+//
+// Have to do this because of the way we are currently bundling the code
+// in this repo, each block gets its own copy of this file.
+//
+// We COULD fix this by doing some webpack magic to spit out the code in
+// `gutenberg-packages` to a shared chunk but assigning `blockTypes` to window
+// is a cheap hack for now that will be fixed once we can merge this code into Gutenberg.
```
michalczaplinski, line 11: In the comment I explain why I'm doing `window.blockTypes` (it's a hack).

src/gutenberg-packages/frontend.js
```diff
@@ -60,4 +107,11 @@ class GutenbergBlock extends HTMLElement {
...
 
-customElements.define( 'gutenberg-interactive-block', GutenbergBlock );
+// We need to wrap the element registration code in a conditional for the same
+// reason we assing `blockTypes` to window (see top of the file).
+//
+// We need to ensure that the component registration code is only run once
+// because it throws if you try to register an element with the same name twice.
+if ( customElements.get( 'gutenberg-interactive-block' ) === undefined ) {
+	customElements.define( 'gutenberg-interactive-block', GutenbergBlock );
+}
```
michalczaplinski, line 117: Directing your attention to the comment 🙂

michalczaplinski:
@luisherranz I can't assign you for review because you've opened the PR 🙂

michalczaplinski:
> There is also an issue that I have mentioned in the video, which I ll document later on. When including the parent block includes only the child block, it does not render that child block. I think this might be an issue with rendering inner blocks that was introduced in https://github.com/luisherranz/block-hydration-experiments/pull/8.

I've managed to debug it and opened a new issue for this: #23

cbravobernal (approved):
Nice job!

src/gutenberg-packages/frontend.js
luisherranz: Yep, this should be part of the Gutenberg APIs, so we can keep the hack for now 🙂

src/blocks/block-hydration-experiments-child/edit.js
```diff
@@ -0,0 +1,19 @@
+// This import is needed to ensure that the `wp.blockEditor` global is available
+// by the time this component gets loaded. The `Title` component consumes the
+// global but cannot import it because it shouldn't be loaded on the frontend of
+// the site.
+import '@wordpress/block-editor';
```
luisherranz, line 5: Can you explain this a little bit? This block doesn't use the `Title` component, does it?

src/gutenberg-packages/frontend.js
```diff
@@ -1,17 +1,37 @@
...
 	if ( !value ) {
 		return null;
 	}
 	return (
 		<gutenberg-inner-blocks
+			ref={( el ) => {
+				if ( el !== null ) {
+					// listen for the ping from the child
+					el.addEventListener( 'gutenberg-context', ( event ) => {
+						event.stopPropagation();
```
luisherranz, line 30: Is there any reason to stop the propagation?

src/gutenberg-packages/frontend.js
```diff
@@ -1,17 +1,37 @@
...
+const Children = ( { value, providedContext } ) => {
 	if ( !value ) {
 		return null;
 	}
 	return (
 		<gutenberg-inner-blocks
+			ref={( el ) => {
+				if ( el !== null ) {
+					// listen for the ping from the child
+					el.addEventListener( 'gutenberg-context', ( event ) => {
```
luisherranz, line 29: I'm not saying this is right or wrong, but what's your reasoning for adding the event listener to the `guntenberg-inner-blocks` instead of the block wrapper? 🙂

luisherranz (commented):
_I can't approve my own PR._

This is fantastic, Michal! 😄👏

I've added a couple of questions, but it's looking great. I'm glad this approach seems to work fine.

If you don't mind, I'm going to rename the blocks as I did in https://github.com/WordPress/block-hydration-experiments/pull/25.

EDIT: I've also moved the `frontend` components to each block folder to make the block structure clearer.

---

The next step would probably be supporting dynamic contexts. I'll add a task for that but basically, we need to make this work:

```js
// Parent provides attributes.count as context
const ParentFrontend = ({ attributes, setAttributes }) => {
  return (
    <button
      onClick={() => {
        setAttributes({ count: attributes.count + 1 });
      }}
    >
      increment context
    </button>
  );
};

// Children is rerendered when context changes (on the parent click)
const ChildrenFrontend = ({ context }) => (
  <div>The count is: {context.count}</div>
);
```

EDIT: I've added a new task for this.

src/blocks/block-hydration-experiments-child/edit.js
michalczaplinski: Oh, it seems that it does work fine even if I remove the `import '@wordpress/block-editor';` which makes sense.

I have added it because it seemed make the warning about `wp.RichText` go away, but looks like I don't need it after all.

src/gutenberg-packages/frontend.js
michalczaplinski: I think that this is the only option we have if I'm not mistaken. Lemme explain in a video:

https://user-images.githubusercontent.com/5417266/169928701-49fedb7d-ff20-4bbb-82e2-364f6d41af26.mp4

src/gutenberg-packages/frontend.js
michalczaplinski: yes! I explain below in the [video](https://github.com/WordPress/block-hydration-experiments/pull/7#discussion_r879974861) 🙂

src/gutenberg-packages/frontend.js
michalczaplinski: I think that this is the only option we have if I'm not mistaken. Lemme explain in a video:

https://user-images.githubusercontent.com/5417266/169930457-e2a19a23-a59c-41ab-963c-56ff84021842.mp4

luisherranz: Ok, thanks for the video, Michal 🙂

I think those problems can be circumvented, but we shouldn't worry for now, this is perfectly fine as it is. If there's a simpler solution, it will present itself once we do more complex stuff!

luisherranz: What would be the benefit of using an extra `div` versus using `this` as you explain in the video?

_`this` points to the `<gutenberg-interactive-block>` wrapper._

michalczaplinski: There would be no benefit, that's what I was trying to say. But the issue is that you only want to ask the parent that is one level up for the context:

I've realized that we could also do this:

```js

// inside of the connectedCallback()

this.addEventListener( 'gutenberg-context', ( event ) => {
  if ( this !== event.target ) { // <- only set the context if we're in the parent. 
    event.stopPropagation();
    event.detail.context = providedContext;
  }
} );

// ping the parent for the context
const event = new CustomEvent( 'gutenberg-context', {
  detail: {},
  bubbles: true,
  cancelable: true,
} );

this.dispatchEvent( event );
```

So, this way we can both listen to and fire the event on the same element (the `<interactive-gutenberg-block>`) but only set the context if we're in the parent. This works because the event bubbles up.

Was this what you were suggesting or did you have another idea in mind?

luisherranz: Yes, I think so 🙂

I think you could also add the event listener after you've dispatched the event so they don't interfere.

PR #8, opened 2022-05-04
Support initially hidden inner blocks

DAreRodz:
This PR adds support for hidden inner blocks.

It duplicates `<InnerBlocks>` inside a `<template>` tag next to the block component. During component hydration, the `<template>` tag is emptied.

### DIFF
--- a/src/gutenberg-packages/frontend.js
+++ b/src/gutenberg-packages/frontend.js
@@ -27,7 +27,7 @@ class GutenbergBlock extends HTMLElement {
       const blockProps = JSON.parse(
         this.getAttribute("data-gutenberg-block-props"),
       );
-      const innerBlocks = this.querySelector("gutenberg-inner-blocks");
+      const innerBlocks = this.querySelector("template.gutenberg-inner-blocks");
       const Comp = blockTypes.get(blockType);
       hydrate(
         <EnvContext.Provider value="frontend">
@@ -41,6 +41,10 @@ class GutenbergBlock extends HTMLElement {
               suppressHydrationWarning={true}
             />
           </Comp>
+          <template
+            className="gutenberg-inner-blocks"
+            suppressHydrationWarning={true}
+          />
         </EnvContext.Provider>,
         this,
       );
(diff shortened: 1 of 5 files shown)


luisherranz:
Exciting! 😄

Any ideas to avoid sending duplicate HTML? We'd need to detect if a block has inner blocks, but they are not in the output. I'm not sure we can solve it here, though. But it'd be nice to start discussing the approach. 

Astro does it in the compiler. My initial guess is that we would need to do it inside the serializer, checking if the rendered block contains `<gutenberg-inner-blocks>` after we [render the markup of the save component](https://github.com/WordPress/gutenberg/blob/trunk/packages/blocks/src/api/serializer.js#L183-L199).

_By the way, I've just added you as a contributor._

luisherranz:
I talked too soon. This is also happening in the `main` branch.

DAreRodz:
I also experienced some problems, but I assumed they came from different block renders after modifying the component. 🤔

luisherranz:
I've debugged it, and it's because of the throw of `useBlockEnvironment`. 

I'll open an issue to inform, but I don't think we need to fix that because it'll never make it to Gutenberg.

So, any ideas to avoid sending duplicate HTML? 🙂

DAreRodz:
I don't have any ideas yet, haha. 😄 I'm trying using the `blocks.getSaveElement` filter, defined [here](https://github.com/WordPress/gutenberg/blob/HEAD/packages/blocks/src/api/serializer.js#L175-L180), but it's not working yet.

DAreRodz:
There are [some docs](https://developer.wordpress.org/block-editor/reference-guides/filters/block-filters/#blocks-getsaveelement) for that filter as well. 👀 

One weird thing I saw is that when you return just the same element you receive in that filter, inner blocks simply disappear. They are not rendered and saved anymore. 🤷

luisherranz:
This is something we shouldn't solve using filters if we ever add this to Gutenberg, so maybe it's not worth trying to solve it here David. It's good to know that `<template>` can be used to populate `children` in the frontend and it works fine in the editor as well.

I'll merge this as it is and we can solve the duplication later 🙂

src/gutenberg-packages/frontend.js
```diff
@@ -41,6 +41,10 @@ class GutenbergBlock extends HTMLElement {
               suppressHydrationWarning={true}
             />
           </Comp>
+          <template
+            className="gutenberg-inner-blocks"
+            suppressHydrationWarning={true}
+          />
```
michalczaplinski, line 47: @DAreRodz Why do we need to render this empty `template` here? If I comment it out, it seems that I can still render initially hidden inner blocks. 🤔

DAreRodz: It's simply to avoid a hydration warning. 😄

michalczaplinski: oh duh, thanks :)

PR #10, opened 2022-05-06
Try autoimport plugin

c4rl0sbr4v0:
⚠️ ⚠️ Experiment

I'm using the same library as NuxtJS in order to try Autoimport features.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "@wordpress/element": "^4.3.0",
     "@wordpress/env": "^4.4.0",
     "@wordpress/scripts": "^22.3.0",
-    "dprint": "^0.24.3"
+    "dprint": "^0.24.3",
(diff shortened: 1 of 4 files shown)


luisherranz:
Thanks, Carlos. This is interesting 🙂

Hmm... we could use something like this to resolve things like `useState` or `RichText` to different things in Edit/Frontend, but I guess that would only work for imports in this whitelist 🤷‍♂️

_I've added you as a collaborator._

c4rl0sbr4v0:
I'm making some other tests in Gutenberg repository to quickly see how much code we could save by this feature.

As you mention, you will need a whitelist (maybe we can include it by default on blocks scaffolding tools), so that any block creator can use @wordpress/element or @wordpress/editor without having to import it from the library.

PR #14, opened 2022-05-10
Implement different hydration techniques

### LINKED ISSUE #12: Implement different hydration techniques
This one should be pretty straightforward.

We could use a `data-gutenberg-hydrate="technique"` with values `load`, `idle` and `view` for now.

Something like:

```js
const technique = this.getAttribute("data-gutenberg-hydrate");

const hydrate = () => {...};

switch (technique) {
  case "idle":
    requestIdleCallback(hydrate)
    break;
  case "view":
    onInteresect(this, hydrate)
    break;
  default:
    hydrate();
}
```

For reference, Astro's code is here: https://github.com/withastro/astro/tree/main/packages/astro/src/runtime/client.

cbravobernal:
Fixes #12 

I added an initial approach to the different hydration techniques as Astro (copy-pasted from the issue and Astro's codebase, mainly).

Should be the technique a prop in block.json rather than an attribute? (I guess choosing the hydration technique on the editor is an overkill)

Should we allow the Block developer to choose the hydration technique or is it a complex concept for most developers? (I still have to check some docs about IntersectionObserver and RequestIdleCallback)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -43,10 +43,10 @@ Feel free to inspect the code, open issues, submit PRs, ask questions...
 - [ ] Avoid bundling unnecessary code (like the Save serializer or lodash) in the Frontend.
 - [ ] Pattern to export different code depending on the context (Edit, Save, or Frontend).
 - [ ] Bundle Preact (compat) instead of React in the Frontend (up for discussion).
-- [ ] Implement different hydration techniques:
-  - [ ] Load
-  - [ ] Idle
-  - [ ] View
-  - [ ] Media
+- [x] Implement different hydration techniques:https://github.com/luisherranz/block-hydration-experiments/pull/14.
+  - [x] Load
+  - [x] Idle
+  - [x] View
+  - [x] Media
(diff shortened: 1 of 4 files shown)


c4rl0sbr4v0:
Do you think we should include media-query-based hydration? [Astro does](https://github.com/withastro/astro/blob/main/packages/astro/src/runtime/client/media.ts), but you have to specify `<SidebarToggle client:media="(max-width: 50em)" />` the media query as an HTML attribute.
Also, [in their documentation ](https://docs.astro.build/en/reference/directives-reference/#clientmedia), they mention:

_If the component is already hidden and shown by a media query in your CSS, then it can be easier to just use client:visible and not pass that same media query into the directive._

In this case, if we don't include the option and just rely in CSS and its visible property, we would be skipping some use cases.

michalczaplinski (approved):
This looks great! 

I think we can merge it as is and continue the conversation about the further details.

luisherranz:
_Just a small note here if/when we propose this in Gutenberg: the intersection observer should also have a fallback._

PR #17, opened 2022-05-17
dprint: Start using spaceAround option

luisherranz:
I was looking at `dprint` PRs to see how difficult it would be to add the WordPress spacing and bumped into [this PR](https://github.com/dprint/dprint-plugin-typescript/pull/341) from a month ago. I'm not sure if it fulfills all WordPress's coding standard needs, but it's a good start.

It's also funny to see how freaking fast this thing is 😄

https://user-images.githubusercontent.com/3305402/168826669-19d5eced-8790-46aa-b6de-e49aaa9169e0.mov


gziolo:
Is there something similar for brackets `[]` and curly brackets `{}`?

As far as I remember, in WordPress, you still would skip the space if there are no params inside `()`.

luisherranz:
> you still would skip the space if there are no params inside ()

Yes, [I've noticed it as well](https://github.com/luisherranz/block-hydration-experiments/pull/17/files#diff-b2337e3a3a6f6e88ff74635ab03fb21d825735ccb5a5128a85e31e020b528402R4), and it's a bit weird:

```js
return useBlockEnvironment(  ) === "edit" ? ...
```

> Is there something similar for brackets [] and curly brackets {}?

Brackets and curly braces are also getting a space with this configuration. Am I missing something?

luisherranz:
> you still would skip the space if there are no params inside ()

I don't see any test for this in [the original PR](https://github.com/dprint/dprint-plugin-typescript/pull/341/files#diff-33ee47b222784b3c9734e47944665c88f6ca98b79e8b5f8b8fa627938808dc50): `testing() => testing(  )`, so maybe this is unexpected. 

It should be easy to fix.

luisherranz:
I've made a PR to fix that: https://github.com/dprint/dprint-plugin-typescript/pull/368 👍

luisherranz:
I've updated `dprint` options to be close to the [JavaScript WordPress coding standards](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/). _I'll explain in the [help test `dprint` issue](https://github.com/luisherranz/block-hydration-experiments/issues/16) what is still missing._

I've also added a GH action that fails if there are any problems with the formatting.

I'm going to merge this now.

PR #18, opened 2022-05-18
Add fallback for intersection observer hydration

cbravobernal:
As mentioned [on #14 comment](https://github.com/luisherranz/block-hydration-experiments/pull/14#issuecomment-1128595462), we should have a fallback for IntersectionObserver, as is not 100% compatible with all browsers.

Screenshot 2022-05-18 at 17 25 41

I just added a try, catch , calling the hydration function if it fails.


luisherranz (approved):
I'm not 100% sure this is the _"standard"_ way to detect if the intersection observer is supported. We've used [this logic](https://github.com/w3c/IntersectionObserver/issues/296#issuecomment-452230176) in the past. But not a big deal, this should be more than ok for our purpose 🙂

PR #26, opened 2022-05-23
Render multiple inner blocks in the save wrapper

### LINKED ISSUE #23: Cannot render `InnerBlocks.Content` more than once
In our `wordpress-blocks.js` we are rendering the `<InnerBlocks.Content/>` twice: 

https://github.com/WordPress/block-hydration-experiments/blob/cb839e18c77a463ebe1c3b504f9f206c9faa81b0/src/gutenberg-packages/wordpress-blocks.js#L21

https://github.com/WordPress/block-hydration-experiments/blob/cb839e18c77a463ebe1c3b504f9f206c9faa81b0/src/gutenberg-packages/wordpress-blocks.js#L29

However, it's [not possible to render `<InnerBlocks/>` or `<InnerBlocks.Content/>` more than once](https://github.com/WordPress/gutenberg/blob/360511303434aa707c507ce621d37971bf08be32/packages/block-editor/src/components/inner-blocks/README.md) !

<a href="https://www.loom.com/share/96f545fad2ad4361b71603618d9570d9">
    
  </a>

https://www.loom.com/share/96f545fad2ad4361b71603618d9570d9

luisherranz:
This is a quick fix for #23, so we can continue experimenting with these APIs.

---

Obviously, we [should not serialize the inner blocks twice](https://github.com/WordPress/block-hydration-experiments/pull/8#issuecomment-1118260158), but we'll improve that if/when we propose this in Gutenberg.

### DIFF
--- a/src/gutenberg-packages/wordpress-blocks.js
+++ b/src/gutenberg-packages/wordpress-blocks.js
@@ -1,4 +1,4 @@
-import { InnerBlocks, useBlockProps } from '@wordpress/block-editor';
+import { useBlockProps, useInnerBlocksProps } from '@wordpress/block-editor';
 import { registerBlockType as gutenbergRegisterBlockType } from '@wordpress/blocks';
 import { getFrontendAttributes } from './utils';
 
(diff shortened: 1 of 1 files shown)


michalczaplinski (approved):
Nice! 🙂 

So, just to be clear, with `useInnerBlocksProps.save();` it is possible to render the inner blocks multiple times after all...?

luisherranz:
I think that technically, you are rendering them only once, but you are adding them to the markup twice.

michalczaplinski:
got it, thanks! 🙁

PR #28, opened 2022-05-24
Support providing context from non-interactive blocks

michalczaplinski:
⚠️ This is still WIP

We have added support for BlockContext in #7. 

Interactive blocks can now provide [block context](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-context/) to other interactive blocks on the frontend. 

However, static (not hydrated) blocks cannot provide context to interactive blocks. This PR is adding support for this feature.


michalczaplinski:
I've recorded a brief overview of what we've been trying to achieve. If anybody would like to continue this work, you're very welcome to do so!


https://user-images.githubusercontent.com/5417266/170140309-fdd1a695-d933-4e13-a0fb-52d84bdd01bb.mp4

michalczaplinski:
Our idea was simply:
- Read the `context` that is passed from the static parent in the `save()` of the interactive block.
- Serialize it in and add it to the attribute like in the `save()`.
- Deserialize and use in the frontend.

But you're right that with the context not being available on `save()` it's not going to work.

> If a block provides a context, the value is in one of its attributes. We just need to serialize that value, then add a global event listener.

Yup, but I think that would require some modification to the "static" blocks, like adding a wrapper around the `save()` of the static blocks and we were trying to avoid that. Unless you had a different idea here? 🤔

luisherranz:
> like adding a wrapper around the save() of the static blocks

I don't really care if it's a wrapper or a `blocks.getSaveContent.extraProps` filter, as long as it is abstracted by Gutenberg and not exposed to the developer 🙂

michalczaplinski:
We paired up to pair-program on this PR with @luisherranz.

We still have to figure out where and how to serialize the data in the static (non-interactive) block:

image

michalczaplinski:
I was able to make it work so far in the simplest case: one static (non-interactive) parent and one child interactive block 👍 .

Most immediate next steps would be:
- Making it work with multiple parents. An interactive block can inherit context from multiple parent blocks.
- Making it work with initially hidden blocks. Adding a global event listener on load does not work because if a new interactive block is dynamically added to the DOM, it will not be subscribed to the events! I guess we have to encapsulate this code in the `connectedCallback()` of our `<gutenberg-interactive-block>` custom component.

michalczaplinski:
I made some more progress:
- Got it to work with multiple parents
- Got it to work with initially hidden blocks

(basically, what I have mentioned in [the comment](https://github.com/WordPress/block-hydration-experiments/pull/28#issuecomment-1175616435) above)

I'm sure it's still buggy but seems to work at least in the happy path.

https://user-images.githubusercontent.com/5417266/177664328-72749558-412b-47d2-9898-50c551edd646.mp4

src/gutenberg-packages/frontend.js
```diff
@@ -27,8 +27,12 @@ const Children = ( { value, providedContext } ) => {
...
 					// listen for the ping from the child
 					el.addEventListener( 'gutenberg-context', ( event ) => {
-						event.stopPropagation();
-						event.detail.context = providedContext;
+						// We have to also destructure `event.detail.context` because there can
+						// already exist a property in the context with the same name.
+						event.detail.context = {
+							...providedContext,
+							...event?.detail?.context,
+						};
```
michalczaplinski, line 35: This is a bit difficult to wrap your head around. The point is that it allows us to "inherit" context across multiple levels of nested blocks.

src/gutenberg-packages/frontend.js
```diff
@@ -114,6 +118,20 @@ class GutenbergBlock extends HTMLElement {
...
 
+class StaticContext extends HTMLElement {
+	connectedCallback() {
+		setTimeout( () => {
+			this.addEventListener( 'gutenberg-context', ( event ) => {
+				const context = JSON.parse( this.attributes.context.value );
+
+				// We have to also destructure `event.detail.context` because there can
+				// already exist a property in the context with the same name.
+				event.detail.context = { ...context, ...event?.detail?.context };
```
michalczaplinski, line 129: Same as in the previous comment: The point is that it allows us to "inherit" context across multiple levels of nested blocks.

src/gutenberg-packages/wordpress-blocks.js
```diff
@@ -21,10 +25,66 @@ const save = ( name, Comp ) =>
...
+export const registerBlockType = (
+	name,
+	{ frontend, save: BlockSave, edit, ...rest },
+) => {
 	gutenbergRegisterBlockType( name, {
 		edit,
-		save: save( name, frontend ),
+		save: BlockSave ?
+			saveWithStaticContext( name, BlockSave ) :
+			save( name, frontend ),
```
michalczaplinski, line 87: I am making an assumption here that:
- Only blocks that have a `frontend.js` file == hydrated interactive blocks
- All blocks that have a `save.js` file == non-interactive blocks that possibly pass block context to other blocks (including interactive hydrated blocks)

michalczaplinski:
A super quick demo of how it works now:

https://user-images.githubusercontent.com/5417266/178624704-7f124cba-8c32-4a79-9890-3b7b7eb7a751.mp4

src/gutenberg-packages/frontend.js
```diff
@@ -114,6 +118,20 @@ class GutenbergBlock extends HTMLElement {
 	}
 }
 
+class StaticContext extends HTMLElement {
+	connectedCallback() {
+		setTimeout( () => {
+			this.addEventListener( 'gutenberg-context', ( event ) => {
```
luisherranz, line 124: I'd rather use a data attribute with a Mutation Observer to avoid introducing extra nodes.

```js
			this.querySelectorAll( '[data-gutenberg-static-context]' ).map( node =>
			 node.addEventListener( 'gutenberg-context', ( event ) => {
```

luisherranz (dismissed):
It looks great! I love how it hooks to the already existing `gutenberg-context` event 🙂

As we discussed, I think it should be better to use a data attribute and a mutation observer to attach the event listeners. It should also be good to add it in PHP instead of JS, so it's not serialized.

src/gutenberg-packages/frontend.js
michalczaplinski: On second thought:
- I worry a bit that putting a MutationObserver at document level might be a performance killer if we run a function on every DOM change in the document.
- It looks like Qwik is not using a MutationObserver? 🤔 I could swear I saw it when we were pair programming, but it's just [using an IntersectionObserver](https://github.com/BuilderIO/qwik/blob/main/packages/qwik/src/qwikloader.ts#L113).

luisherranz: > a MutationObserver at document level might be a performance killer

Could you investigate that? I'm not too familiar with Mutation Observers, to be honest.

michalczaplinski: All right, I ll try to change the implementation to use the MutationObserver and create a component that does high frequency DOM updates to see how bad the performance can get.

My hunch is that it won't be great because there can be multiple DOM updates per second and coming from browser extensions or ads that we can't control.

luisherranz: Try using [attributeFilter](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe#using_attributefilter).

```js
const observer = new MutationObserver(() => {
  // Only triggered when a `data-gutenberg-static-context` attribute changes
});

observer.observe(document, {
  attributes: true,
  attributeFilter: ['data-gutenberg-static-context'],
  subtree: true
});
```

michalczaplinski: oh, I didn't know about this! Yeah, that might just work 🙂

luisherranz:
I'm resolving the conflicts with my refactor on https://github.com/WordPress/block-hydration-experiments/pull/40/commits/d61391cf240e82b5edafbd8f3af081f41331318d.

luisherranz:
@michalczaplinski everything should be working again now. Feel free to continue with this 🙂

src/gutenberg-packages/frontend.js
luisherranz: @michalczaplinski, I'll try to take a look at this today! I hope you don't mind 🙂

luisherranz: Sadly `attributeFilter` only works when an attribute is modified, not when an element with that attribute is added to the DOM.

I'm going to keep your initial custom-element implementation, although I'll add it dynamically from PHP to avoid unnecessary serialization.

It has the caveat that it won't support sourced attributes, though.

luisherranz: I've done something a bit different. I've merged both `static-block` and `gutenberg-interactive-block` in `gutenberg-block`, and only hydrate blocks containing `data-gutenberg-hydration`. That way, no extra code is required for non-interactive blocks.

luisherranz:
As the Mutation Observer option [could introduce performance concerns](https://github.com/WordPress/block-hydration-experiments/pull/28#discussion_r921580868) and the [attributeFilter](https://github.com/WordPress/block-hydration-experiments/pull/28#discussion_r928776197) approach didn't work, I just renamed `gutenberg-interactive-block` to `gutenberg-block` and reused all the code.

I've also renamed all the blocks. I hope the new names will help to make more sense of the structure of these experiments for new people looking into them.

I've made a video to explain the changes.

<a href="https://www.loom.com/share/72d45ab8574c42258e8eee48c3dca3d5">
    
  </a>

https://www.loom.com/share/72d45ab8574c42258e8eee48c3dca3d5

@michalczaplinski would you mind taking a look? Thanks!! 🙂

luisherranz:
I'm going to resolve the conflicts with https://github.com/WordPress/block-hydration-experiments/pull/41.

luisherranz:
Done 👍️

AFAIK, everything should be ready to merge now.

michalczaplinski:
Looks great, thanks for working on this, @luisherranz ! 🙂

I can't approve my own PR just going to merge it.

luisherranz:
Thanks, Michal! 🎉

PR #32, opened 2022-05-30
fix rendering of applied rich text formats in proxy component

fabiankaegy:
The value that gets passed into the `RichText`component is a text string that may include actual markup in cases where rich text formats like bold / italics etc are applied to the string. In order for them to render correctly, we need to set the text as inner HTML.


luisherranz (approved):
Thanks for the fix Fabian 🙂

PR #34, opened 2022-06-14
Add the interactive block attributes during server-side rendering

### LINKED ISSUE #33: Add the interactive block attributes during server-side rendering
Ok, so we've ended up serializing a lot of stuff in the markup:

```html
<gutenberg-interactive-block
  data-gunteberg-block-type="..."
  data-gunteberg-context-used="..."
  data-gunteberg-context-provided="..."
  data-gunteberg-context-attributes="..."
  data-gunteberg-context-sourced-attributes="..."
  data-gunteberg-context-block-props="..."
  data-gunteberg-context-hydrate="..."
>
  <!-- The real block -->
  <div class="...">...</div>
</gutenberg-interactive-block>
```

The saved markup doesn't seem the best place for this, so I wonder if we should add all those attributes dynamically during the server-side rendering.

Any thoughts? 🙂

ockham:
WIP. Aims to fix #33.

TODO (?)
- [ ] `blockProps`
- [ ] `innerBlockProps`

### DIFF
--- a/block-hydration-experiments.php
+++ b/block-hydration-experiments.php
@@ -14,3 +14,38 @@ function block_hydration_experiments_init() {
 	register_block_type( plugin_dir_path( __FILE__ ) . 'build/blocks/block-hydration-experiments-parent/' );
 }
 add_action( 'init', 'block_hydration_experiments_init' );
+
+function bhe_block_wrapper( $block_content, $block, $instance ) {
+	// We might want to use a flag from block.json as the criterion here.
+	if ( ! in_array(
+		$block['blockName'],
+		array(
+			'bhe/block-hydration-experiments-parent',
+			'bhe/block-hydration-experiments-child'
+		),
+		true
+	) ) {
+		return $block_content;
+	}
+
+	$block_wrapper = sprintf(
+		'<gutenberg-interactive-block ' .
+		'data-gutenberg-block-type="%1$s" ' .
+		'data-gutenberg-context-used="%2$s" ' .
+		'data-gutenberg-context-provided="%3$s" ' .
+		'data-gutenberg-attributes="%4$s" ' .
+		'data-gutenberg-hydrate="idle">',
+		esc_attr( $block['blockName'] ),
+		esc_attr( json_encode( $instance->block_type->uses_context ) ),
+		esc_attr( json_encode( $instance->block_type->provides_context ) ),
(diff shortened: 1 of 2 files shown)


ockham:
During our pair programming session today, @luisherranz and I were experimenting with how to best rehydrate `blockProps`.

It might be a bit harder to extract them (reliably) on the server side; it might be possible by basically invoking server-side block-supports functionality, but we're not 100% sure if this covers all client-side block props (especially if added by 3rd party plugins).

We thus tried to extract block props [from the DOM](https://github.com/WordPress/block-hydration-experiments/blob/4631ad7484d10c5cd122fc3221823c73425fc80f/src/blocks/block-hydration-experiments-parent/frontend/index.js#L12). We know that they only affect `className` and `style`, and are typically applied to the outermost DOM element. We thus tried something like

```diff
diff --git a/src/gutenberg-packages/frontend.js b/src/gutenberg-packages/frontend.js
index 4446742..787519e 100644
--- a/src/gutenberg-packages/frontend.js
+++ b/src/gutenberg-packages/frontend.js
@@ -78,9 +78,11 @@ class GutenbergBlock extends HTMLElement {
                        const context = pickKeys( event.detail.context, usesContext );
 
                        const blockType = this.getAttribute( 'data-gutenberg-block-type' );
-                       const blockProps = JSON.parse(
-                               this.getAttribute( 'data-gutenberg-block-props' ),
-                       );
+                       const blockProps = {
+                               className: this.firstChild.className,
+                               style: this.firstChild.style?.cssText
+                       };
+
                        const innerBlocks = this.querySelector(
                                'template.gutenberg-inner-blocks',
                        );
```

This needs a bit more polish: `this.firstChild` doesn't seem to be the wrapping `<div />` but some `#text` node, and React expects a `style` object (with attributes in `camelCase`) rather than a string (as returned by [`.cssText`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/cssText)).

I'll continue to tweak this to get it to work; I might also still try the server-side approach.

---

Finally, note that block authors might choose to apply block props to another element (instead of the outermost wrapper), in which case the above won't work. However, re-hydration of block props is fortunately only really needed if those can be changed interactively. Since it only affects `className` and `style`, this might not be a very large group of use cases.

ockham:
Should be ready for review, I think.

PR #40, opened 2022-07-04
Support React Context API

### LINKED ISSUE #38: Support React's Context API between different blocks
### What
When we asked plugin developers what we should include in our experiments, they answered React Context support. 

Context provides a way to pass data through the component tree without having to pass props down manually at every level, this means easy data communication between different components.

### The problem
In this experiment, we are taking the Island's architecture as the main approach. We use custom elements so we hydrate the block automatically when joins the tree (you can copy-paste a block, and will be hydrated), and also allow us cool things like defining the way we want to hydrate it.

The problem with this approach is, that, usually when you use React, you only have **one application** where you just add Providers components in the tree at the highest level possible, so that way any child component can access the Context data.

In our case, we have multiple **React Apps**, so we have to find a process to share that context between different apps while reusing the React Context API as much as possible. We don't want the developer to learn anything new if they already know how to work with React Context.

### One possible exploration
 
Screenshot 2022-06-28 at 10 29 49

https://excalidraw.com/#json=x8cpUygy5LrBMMzhDmS8E,s5RQCGHTZXZStvCFPDPfwA

On the block registration, we should be able to import the Context, create a wrapper with that Context so our Child Component (the block itself) could access it, and then set a Consumer that can sync the value with the parent Provider, so the Context is updated when needed in each parent.

We thought about the idea of creating a subscriber to a Context, but, in this case, if we have more than one context, is really difficult to select the Context to subscribe to. Instead, we want to try the default behavior of the Context of subscribing to the closest parent Provider, by using a custom event, similar to what we do for the UsesContext and ProvidesContext in default blocks.

We created a video to try to explain everything, recommended playback speed is 1.5x, as I talked toooo slow 😓 

https://user-images.githubusercontent.com/37012961/176136129-dcfec652-fe04-487a-abaf-df7c5a8901af.mp4

c4rl0sbr4v0:
Fixes #38


src/gutenberg-packages/frontend.js
```diff
@@ -84,28 +88,69 @@ class GutenbergBlock extends HTMLElement {
...
 			);
-			const Comp = window.blockTypes.get( blockType );
+			const { Component, options } = window.blockTypes.get( blockType );
+			if ( options?.providesContext?.length > 0 ) {
+				// Refactor to include a forEach
+				this.addEventListener( 'react-context', ( event ) => {
+					// we compare provided and used context
+					if ( event.detail.context === options.providesContext[0] ) {
+						console.log( 'SYNC!!' );
+					}
```
c4rl0sbr4v0, line 98: This means that the ProvidedContext variable and UsesContext variable are the same. We dispatch the UsesContext variable from the child (event.detail.context) to the parent.

We have yet to discover how sync the value of the parent provider with the child block.

src/gutenberg-packages/frontend.js
```diff
@@ -84,28 +88,69 @@ class GutenbergBlock extends HTMLElement {
...
 				<EnvContext.Provider value='frontend'>
-					<Comp
-						attributes={attributes}
-						blockProps={blockProps}
-						suppressHydrationWarning={true}
-						context={context}
+					<ConditionalWrapper
+						condition={UsedReactContext !== undefined}
+						wrapper={children => (
+							<UsedReactContext.Provider value={'to_sync'}>
```
c4rl0sbr4v0, line 120: This value should be given by the Parent provider (which is in another block).

luisherranz:
Initial sync is working, but there are some remaining tasks:

- [ ] Make sure `subscribers` map is only created once
- [ ] Subscribe to block + Context (maybe we could use `this`)
- [ ] Sync initial value as well
- [ ] Check if the React warning of updating `Provider` during the `Consumer` rendering (fixed with the `setTimeout` at the moment) can be avoided
- [ ] Unsubscribe on the `useEffect` return
- [ ] Add `maps` or `forEach` in `usesContext` and `providesContext` arrays
- [ ] Investigate if using `WeakMap` instead of `Map` makes sense to avoid memory leaks

Finally, it'd be good to start thinking about how React Context should be used in the Editor.

c4rl0sbr4v0:
I'm not a memory leaks expert but, with the debug tool, it seems that using WeakMap is a better option, maybe due to its garbage collection behavior.

Without Weakmaps:
without_weakmaps


With Weakmaps
with_weakmaps

c4rl0sbr4v0:
Now the initial value is synced, by using a similar approach as the subscriber.

c4rl0sbr4v0:
We also have to test if it works with different providers.

luisherranz:
This is ready now.

I've done a big refactoring of `frontend.js` because it was getting a bit out of hand. There will be merge conflicts in https://github.com/WordPress/block-hydration-experiments/pull/28, but I think it was worth it.

I've also removed dprint because it was failing, and I didn't know why and I wanted to remove it anyway.

PR #41, opened 2022-07-04
Support className and style blockProps on block wrapper

### LINKED ISSUE #36: Support block support class names and inline styles on the block wrapper and make `blockProps` optional
In #3, @luisherranz wrote

> Block supports (only in the wrapper node, no skip-serialization): I'm serializing the block props (which include the `has-text-color`... class names) because I thought React would remove them, but it doesn't seem to do so. I need to investigate more.

"Serializing the block props" means that they're [passed as a `data-gutenberg-block-props` attribute](https://github.com/WordPress/block-hydration-experiments/blob/4631ad7484d10c5cd122fc3221823c73425fc80f/src/gutenberg-packages/frontend.js#L81-L83), and [rehydrated](https://github.com/WordPress/block-hydration-experiments/blob/4631ad7484d10c5cd122fc3221823c73425fc80f/src/gutenberg-packages/frontend.js#L95) based on that, rather than relying on React to take care of the rehydration.

<details>
<summary>
Diff to disable this "manual" rehydration.
</summary>

```diff
diff --git a/src/gutenberg-packages/frontend.js b/src/gutenberg-packages/frontend.js
index 4446742..e0fbe42 100644
--- a/src/gutenberg-packages/frontend.js
+++ b/src/gutenberg-packages/frontend.js
@@ -92,7 +92,7 @@ class GutenbergBlock extends HTMLElement {
                                <EnvContext.Provider value='frontend'>
                                        <Comp
                                                attributes={attributes}
-                                               blockProps={blockProps}
+                                               //blockProps={blockProps}
                                                suppressHydrationWarning={true}
                                                context={context}
                                        >
```
</details>

During [our pair-programming session today](https://github.com/WordPress/block-hydration-experiments/pull/34#issuecomment-1156699778), we actually found that having props of the same name react to interaction causes React to remove the original block props:

```diff
--- a/src/blocks/block-hydration-experiments-parent/frontend/index.js
+++ b/src/blocks/block-hydration-experiments-parent/frontend/index.js
@@ -9,7 +9,7 @@ const Frontend = (
        const [ counter, setCounter ] = useState( initialCounter );
 
        return (
-               <div {...blockProps}>
+               <div className={ show ? 'shown' : 'hidden' } {...blockProps}>
                        <Title message={message} />
                        <Button handler={() => setShow( !show )} />
                        <button onClick={() => setCounter( counter + 1 )}>{counter}</button>
```

Thus, with both diffs applied:

broken-hydration

ockham:
Fixes #36.

This is the result from a pair-programming session with @luisherranz.

Use the server-side implementation of the block-supports mechanism to communicate block props as `data-` HTML attributes to the interactive Gutenberg block (custom element) on the frontend.

For an alternative approach we tried earlier (but that didn't work), see https://github.com/WordPress/block-hydration-experiments/issues/36#issuecomment-1174143198.

TODO:
- [x] Make sure `style` blockProps also work.

block-props-hydration-working


luisherranz:
I'm resolving the conflicts with my refactor on https://github.com/WordPress/block-hydration-experiments/commit/d61391cf240e82b5edafbd8f3af081f41331318d.

luisherranz:
I think everything should be fine again.

ockham:
Okay, I think I managed to figure out dynamic setting of `style`.

(It's pretty much like we thought it should work during our pairing. I _think_ that we'd gotten stuck due to poor choice of values for `fontWeight`; turns out that [those can't be `> 1000`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight), and it seems that React is smart enough to ignore any such values, instead keeping the previous one 😅 )

luisherranz (approved):
LGTM. Thanks @ockham!! 🙂👏

Issue #37, opened 2022-06-22
Nomenclature suggestions


mtias:
Great to see all the explorations and progress. Even though nothing is finalized yet, it seems timely to look at the API surface and some of the naming conventions to ensure we are putting forward the best possible set of tools.

Can we consider whether `View` is a better nomenclature to use than `Frontend`?

```js
registerBlockType( 'block-name', {
	edit: Edit,
	// The Save component is derived from the View component.
	view: View,
} );
```

and

```js
import View from './view';
registerBlockType( 'block-name', View );
```

---

### Registration

```js
// Before
class GutenbergBlock extends HTMLElement {}

// After
class WpBlock extends HTMLElement {}
```

Same goes for other APIs like events and attributes:

```js
// Before
new CustomEvent( 'gutenberg-context', {} )

// After
new CustomEvent( 'block-context', {} )
// or
new CustomEvent( 'wp-block-context', {} )
```

Similarly `data-` attributes should use `block` or `wp-block` instead of `gutenberg` namespace.

---

### Custom Elements

Super interesting to see the use of custom elements, I think it's a good path forward. I'd suggest improving the names as well:

```html
<!-- before -->
<gutenberg-interactive-block>

<!-- after -->
<wp-block>
```

Can we assume interactivity if it's declared as a custom element?

```html
<!-- before -->
<gutenberg-inner-blocks>

<!-- after -->
<wp-inner-blocks>
```

luisherranz:
Hey Matías, thanks for chiming in because, as far as I know, most of the contributors collaborating here are not familiar with Gutenberg nomenclature yet (me the first!), so the sooner we learn about it, the better 🙂

> Can we consider whether `View` is a better nomenclature to use than `Frontend`

Sure. I used the `frontend` name because there is a possibility that we want to differentiate between the current `view` script enqueuing behavior, which is always enqueued, and an enqueuing that is deferred in time based on the different [hydration technique](https://github.com/WordPress/block-hydration-experiments/issues/12), which may need a different name. But it's not even clear if we won't be able to do so with backward compatibility with `view`, so I agree to keep the `view` name for now. We'll make a PR to change it 🙂

> `wp-block` vs `block-` instead of `gutenberg-`

This has been more of a surprise to me. 

So it looks like **Gutenberg** is only the project name, but none of the APIs should have the Gutenberg name. Is that correct?

[...]

PR #45, opened 2022-07-26
Nomenclature changes

luisherranz:
Fixes https://github.com/WordPress/block-hydration-experiments/issues/37

- [x] Replace `frontend.js` files and `Frontend` componentes with `view.js` and `View`
- [x] Replace `"frontend": true` attribute prop with `"public": true`
- [x] Replace `gutenberg-` with `wp-block-`

### DIFF
--- a/block-hydration-experiments.php
+++ b/block-hydration-experiments.php
@@ -80,14 +80,14 @@ function bhe_block_wrapper($block_content, $block, $instance)
 
 	$block_wrapper =
 		sprintf(
-			'<gutenberg-block ' .
-				'data-gutenberg-block-type="%1$s" ' .
-				'data-gutenberg-uses-block-context="%2$s" ' .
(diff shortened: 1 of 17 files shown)


michalczaplinski (approved):
I'm not crazy about those `wp-` prefixes but for now it's not a problem if we want to change them in the future. We should just be mindful so that this "internal" name does not get stuck like FSE (Full Site Editing) did 🙂 .

ockham:
> I'm not crazy about those `wp-` prefixes but for now it's not a problem if we want to change them in the future. We should just be mindful so that this "internal" name does not get stuck like FSE (Full Site Editing) did 🙂 .

Reading #37, it seems to be that the whole purpose of _this_ PR is to avoid sticking with a `gutenberg-` prefix (in favor of `wp-`) 🙈

Issue #42, opened 2022-07-14
Unmount interactive blocks before removing them from the DOM


luisherranz:
While working on the React Context API, @c4rl0sbr4v0 discovered that with the current implementation, the unmount callbacks (mainly returns of `useEffect`) won't run when an interactive block is removed from the DOM because a parent interactive block stops showing `children`. This is normal because they are independent React apps, but can lead to memory leaks and inconsistencies.

I've taken a look at Astro, and [it doesn't support this](https://stackblitz.com/edit/astro-v5s1nn?file=src%2Fcomponents%2FChildren.jsx). I couldn't find any reference to this problem in their repository either. I haven't looked into other island-based frameworks.

I've been playing around with this, and the callback of `useLayoutEffect` runs before the DOM nodes are removed, so maybe we could use that to unmount the children interactive blocks before they are destroyed.

```js
const Children = ({ value, providedContext }) => {
  const ref = useRef(null);

  useLayoutEffect(
    () => () => {
      ref.current
        .querySelectorAll("gutenberg-interactive-block")
        .map((node) => {
          // These interactive blocks are about to be destroyed. We can send them a
          // signal so they can switch to `null` internally and force React to run
          // their unmount callbacks.
          const setIsMounted = isMountedMap.get(node);
          setIsMounted(false);
        });
    },
    []
  );

  return (
    <gutenberg-inner-blocks
      ref={ref}
      suppressHydrationWarning={true}
      dangerouslySetInnerHTML={{ __html: value }}
    />
  );
};
```

The other part is to switch to `null` if a node receives that signal. We can use a similar approach to the one we used in the React Context.

```js
const InteractiveBlock = () => {
  const [isMounted, setIsMounted] = useState(true);

  useEffect(() => {
    // Expose the setIsMounted so it can be accessed by the parent.
    isMountedMap.set(this, setIsMounted);
  })

  return isMounted ? <Comp ... /> : null;
};

hydrate(InteractiveBlock, this);
```

I'm not 100% confident that React will successfully run all the callbacks with this approach, but I don't consider this a priority, especially considering that even Astro doesn't have a solution for this. It'd be great to keep an eye on it, though.

EDIT: Client-side transitions could be another source of external unmounting. There could be more. Maybe the API should be part of the HTML nodes? Something like [Bento Component APIs](https://bentojs.dev/components/bento-accordion/):

Bento:

```js
const accordion = document.querySelector("#myAccordion");
[...]

PR #47, opened 2022-08-03
Unmount react component when `WpBlock` is disconnected

DAreRodz:
Fixes https://github.com/WordPress/block-hydration-experiments/issues/42.

The solution is straightforward. Inside the [`disconnectCallback()`](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#using_the_lifecycle_callbacks) defined in `WpBlock`, we run [`unmountComponentAtNode(this)`](https://reactjs.org/docs/react-dom.html#unmountcomponentatnode) to safely unmount the React tree within the `WpBlock` instance.

Here is a video where I show how it works: https://github.com/WordPress/block-hydration-experiments/issues/42#issuecomment-1203271407

### DIFF
--- a/src/gutenberg-packages/frontend.js
+++ b/src/gutenberg-packages/frontend.js
@@ -1,6 +1,7 @@
 import { Consumer, createProvider } from './react-context';
 import { createGlobal, matcherFromSource } from './utils';
 import { EnvContext, hydrate } from './wordpress-element';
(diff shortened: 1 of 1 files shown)


DAreRodz:
I don't think this really needs a review, hehe, it's already tested and works fine. I'll just merge it. 🚀

PR #48, opened 2022-08-08
Use block support flag for hydration opt-in

yscik:
Change the hard-coded list of supported blocks in the `render_block` hook. 
Use a support flag, `view`, to enable frontend block hydration.

Also use an extend form of the flag to control the hydration technique setting. 

At this point this is just a quick step to allow adding new blocks, I imagine the final API might look different.

### DIFF
--- a/block-hydration-experiments.php
+++ b/block-hydration-experiments.php
@@ -25,22 +25,12 @@ add_action('init', 'block_hydration_experiments_init');
 
 function bhe_block_wrapper($block_content, $block, $instance)
 {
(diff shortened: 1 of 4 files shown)


luisherranz (approved):
Yeah, we need to start thinking about the opt-in mechanism 🙂

I don't like `view` that much, as [that name is already used by people using `viewScript` to enqueue regular JS files](https://github.com/WordPress/block-hydration-experiments/issues/37#issuecomment-1164566521) and I think it can be confusing, but we can figure that out later.

Feel free to merge 👍

Issue #52, opened 2022-08-09
Make sure View components are automatically hydrated even if their component is registered after the connectedCallback execution


luisherranz:
In order to make this hydration technique as solid as possible, it is important that it doesn't rely on any execution order. One of our goals is to **support client-side transitions** where both the new HTML (containing `wp-block` tags) and the JS files associated with those blocks can come in any other, including the DOM modification first (and therefore the `connectedCallback` execution) and the frontend component (JS file) later. We need to make sure that hydration works even if the frontend component is registered after the `connectedCallback` execution.

This doesn't need to be complicated. I think that if some `blockType` is missing from the registry, the `connectedCallback` can just annotate it somehow so the registry `registerBlockType()` can trigger the hydration manually when it receives that `blockType`.

Something like this:

```js
connectedCallback() {
  if (blockTypes.has(blockType)) {
    hydrate({ blockType, element: this, ...});
  } else {
    // ... anotate for latter hydration
    blockTypesToHydrate.set(name, { element, ... });
  };
}
```

```js
const registerBlockType = (name, Component, options) => {
  blockTypes.set(name, { Component, options });

  // Hydrate annotated blocks
  if (blockTypesToHydrate.has(name)) {
    const { element, ... } = blockTypesToHydrate.get(name);
    hydrate({ blockType, element, ...});
  }
};
```

---

_Another thing that could cause a race condition could be the JS containing `registerBlockType()` and the components using it, but we can figure out if we need something else for that later._

PR #56, opened 2022-08-18
Rename duplicated register block type frontend function

cbravobernal:
Fixes #55 

Right now, we are using the same name for two different functions that do different things, this may cause understanding errors.

[In wordpress-blocks.js file](https://github.com/WordPress/block-hydration-experiments/blob/2d8a9ff55314e6f96a23aac535f1a91c944da213/src/gutenberg-packages/wordpress-blocks.js#L21):
```
export const registerBlockType = (name, { edit, view, ...rest }) => {
	gutenbergRegisterBlockType(name, {
		edit,
		save: Wrapper(view),
		...rest,
	});
};
```

[In gutenberg-packages/frontend.js file](https://github.com/WordPress/block-hydration-experiments/blob/2d8a9ff55314e6f96a23aac535f1a91c944da213/src/gutenberg-packages/frontend.js#L8)
```
export const registerBlockType = (name, Component, options) => {
	blockTypes.set(name, { Component, options });
};
```

After a little chat, we think that add the suffix Frontend will make things easier.

PR #57, opened 2022-08-24
Hydrate views even if they are registered after the DOM nodes appear

luisherranz:
@darerodz and I did a pair programming this morning that resulted in this PR to fix https://github.com/WordPress/block-hydration-experiments/issues/52. 

## What

We changed the logic to make sure there are no race conditions between the `View` components being registered and the custom elements' `connectedCallback` execution.

## Why

Because sometimes the registration will happen after the `connectedCallback` execution.

## How

Create a global registry that can be accessed by any party, and delay the hydration of the custom elements that have no `View` component registered yet.

<a href="https://www.loom.com/share/e01d077aa2c74bf5a0b59a606dee28dc">
    
  </a>

https://www.loom.com/share/e01d077aa2c74bf5a0b59a606dee28dc

---

## Additional comments

I've also renamed `registerBlockType` to `registerBlockView` as an alternative to https://github.com/WordPress/block-hydration-experiments/pull/56 and removed all references to the word "frontend", even in file names.

cc: @c4rl0sbr4v0: would you mind taking a look at that? Thanks!


src/gutenberg-packages/utils.js
```diff
@@ -9,12 +9,11 @@ export const matcherFromSource = (sourceConfig) => {
...
-// this repo, each block gets its own copy of this file, but this won't happen
-// if/when we do this in Gutenberg.
 export const createGlobal = (name, initialValue) => {
-	if (typeof window[name] === 'undefined') {
-		window[name] = initialValue;
-	}
-	return window[name];
+	['wp', 'view', name].reduce((obj, name, i) => {
+		if (typeof obj[name] === 'undefined')
+			obj[name] = i === 2 ? initialValue : {};
```
DAreRodz, line 15: Interesting way to initialize a nested property inside an undefined object. 🙂

I would use the length from the array instead of hardcoding it to check if we have reached the property with the initial value:

```suggestion
	['wp', 'view', name].reduce((obj, name, i, array) => {
		if (typeof obj[name] === 'undefined')
			obj[name] = i === (array.length - 1) ? initialValue : {};
```

DAreRodz (approved):
Nice implementation and refactoring, Luis. 👏

src/gutenberg-packages/utils.js
luisherranz: Yeah, but why calculate `array.length - 1` when it is always 2 anyway. And if someone changes the array length in the future and doesn't update the hardcoded value, this thing won't work at all. So I think it's fine with the hardcoded 2 😄🤷‍♂️

luisherranz: I'm going to merge this as it is, but feel free to refute my argument!

PR #61, opened 2022-08-29
Small animation to make sure we are hydrating correctly

luisherranz:
I've added a small animation to make sure we are hydrating correctly. It should always be smooth. If it jumps to the beginning, it means React destroyed the HTML node and recreated it again.

 
Screen Capture on 2022-08-29 at 17-03-28


luisherranz:
By the way, I added it because hydration can be affected by things like spaces and line breaks.

For example, after prettifying the `index.html` file in [this codesandbox](https://codesandbox.io/s/su6zl1?file=/public/index.html), the hydration fails.

cbravobernal (approved):
LGTM!

Issue #39, opened 2022-06-30
Tracking issue: Custom Elements Hydration 🧩


luisherranz:
This issue aims to list and keep track of the tasks related to the Custom Elements Hydration experiment. 

_For the Directives Hydration experiment, please check the [Tracking Issue: Directives Hydration ⚛️](https://github.com/WordPress/block-hydration-experiments/issues/64)._

The main branch of this experiment is `main-custom-elements-hydration`. Make sure you select it when opening a pull request if it is related to this experiment.

Please proactively comment on this issue whenever you start working on a new task, get blocked, or finish a task, sharing as much detail and information as possible. Thanks!!

# Done

- Hydrate Frontend components
- Make sure Frontend components are automatically rehydrated if they appear in the DOM at any point (not only on page load)
- Support partial hydration with Inner blocks (children raw HTML)
- Serialize attributes and pass them down to the View component
- Use children instead of in the Save component to be able to reuse the same component in the View
- https://github.com/WordPress/block-hydration-experiments/pull/15
- https://github.com/luisherranz/block-hydration-experiments/pull/8
- https://github.com/luisherranz/block-hydration-experiments/pull/3
- https://github.com/luisherranz/block-hydration-experiments/pull/2
- https://github.com/luisherranz/block-hydration-experiments/pull/3
- https://github.com/luisherranz/block-hydration-experiments/pull/14
- https://github.com/luisherranz/block-hydration-experiments/pull/7
- https://github.com/WordPress/block-hydration-experiments/pull/34
- https://github.com/WordPress/block-hydration-experiments/pull/30
- https://github.com/WordPress/block-hydration-experiments/pull/40
- https://github.com/WordPress/block-hydration-experiments/pull/28
- https://github.com/WordPress/block-hydration-experiments/pull/41
- https://github.com/WordPress/block-hydration-experiments/pull/45
- https://github.com/WordPress/block-hydration-experiments/pull/47
- https://github.com/WordPress/block-hydration-experiments/pull/57

# Required to expose as experimental API in Gutenberg

- https://github.com/WordPress/block-hydration-experiments/issues/49
- https://github.com/WordPress/block-hydration-experiments/issues/11
- https://github.com/WordPress/block-hydration-experiments/issues/50
- https://github.com/WordPress/block-hydration-experiments/issues/59
- https://github.com/WordPress/block-hydration-experiments/issues/69

# Future improvements

- https://github.com/WordPress/block-hydration-experiments/issues/62
[...]

PR #65, opened 2022-09-06
⚛️ Restructure the project folder and files to make it ready for the Directives Hydration

luisherranz:
Co-authored with @darerodz.

These are the things we did here:

- Clean a bit the code related to the [Custom Elements Hydration](https://github.com/WordPress/block-hydration-experiments/issues/39)
- Create a separate config for Webpack for the hydration code and the View components
- Enqueue the hydration and View components
- Switch to Preact (compat) on that new Webpack config
- Remove the shared code from the Editor to avoid any `@wordpress` import
  In doing so, we broke the code of the Editor, but we'll fix that in an upcoming PR

The folder/files should be ready now to add the first implementation of the Directives Hydration.


DAreRodz (approved):
LGTM! ✅

Issue #64, opened 2022-09-06
Tracking issue: Directives Hydration ⚛️


luisherranz:
**We are now actively working on the [WordPress Directives Plugin Tracking issue](https://github.com/WordPress/block-hydration-experiments/issues/80) instead** ➡️

---

This issue aims to list and keep track of the tasks related to the **Directives Hydration experiment**.

- https://github.com/WordPress/block-hydration-experiments/issues/60

_For the Custom Elements Hydration experiment, please check the [Tracking Issue: Custom Elements Hydration 🧩](https://github.com/WordPress/block-hydration-experiments/issues/39)._

The main branch of this experiment is `main-directives-hydration`. Make sure you select it when opening a pull request if it is related to this experiment.

Please, proactively comment on this issue whenever you start working on a new task, get blocked, or finish a task, sharing as much detail and information as possible. Thanks!!

# Experiments

- https://github.com/WordPress/gutenberg/pull/44034
- https://github.com/woocommerce/woocommerce-blocks/pull/7187
- https://github.com/WordPress/gutenberg/pull/44509
- https://github.com/DAreRodz/js-framework-benchmark/pull/1

# Done

- https://github.com/WordPress/block-hydration-experiments/pull/65
- https://github.com/WordPress/block-hydration-experiments/pull/66
- https://github.com/WordPress/block-hydration-experiments/pull/68
- https://github.com/WordPress/block-hydration-experiments/pull/70
- https://github.com/WordPress/block-hydration-experiments/pull/71
- https://github.com/WordPress/block-hydration-experiments/pull/72
- https://github.com/WordPress/block-hydration-experiments/pull/77
- https://github.com/WordPress/block-hydration-experiments/pull/74
- https://github.com/WordPress/block-hydration-experiments/pull/78

# Next

# On hold
- https://github.com/WordPress/block-hydration-experiments/issues/75

luisherranz:
I've opened the first PR to prepare the project to add the initial version of the hydration (co-authored with @darerodz):

- https://github.com/WordPress/block-hydration-experiments/pull/65

DAreRodz:
#65 is approved and merged. 🎉  I'll update the tracking issue's description to reflect that.

Aso, I've started with a PR to hydrate block components inside a single vDOM tree:

- https://github.com/WordPress/block-hydration-experiments/pull/66

luisherranz:
I've made a proof of concept of the vdom-based client-side navigation on the Query Loop block:

[...]

PR #66, opened 2022-09-07
⚛️ Hydrate the blocks with Directives Hydration (using `wp-block` and `wp-inner-blocks`)

DAreRodz:
Tracking issue: https://github.com/WordPress/block-hydration-experiments/issues/64

This PR uses the `toVdom()` function from [preact-markup](https://github.com/developit/preact-markup) to hydrate blocks inside a single vDOM tree.

It uses `<wp-block>` and `<wp-inner-blocks>` tags to detect both block components and children.

- [x]  Fix preact hooks
- [x]  Get block props from `data-wp-block-props`
- [x]  Get attributes from `data-wp-block-attributes`
- [x]  Handle sourced attributes defined in `data-wp-block-sourced-attributes`

I'm planning to implement these directives in a different PR, to test how it could be done with the [preact option hooks API](https://preactjs.com/guide/v10/options/).
- Expose attributes defined in `data-wp-block-provides-block-context`
- Consume attributes defined in `data-wp-block-uses-block-context`


DAreRodz:
For context, using components from different entry points was failing because each entry point had its own instance of `preact/hooks`, which needs to be a singleton to work correctly. This is fixed in https://github.com/WordPress/block-hydration-experiments/pull/66/commits/99635c4c9e77d224bd46df41bf134858116c5dea by specifying [`runtimeChunk`](https://webpack.js.org/configuration/optimization/#optimizationruntimechunk) (props to @luisherranz). 🎉

DAreRodz:
I consider this PR ready to review. There is a problem with hydration, though, that it is not working correctly (elements are being regenerated). We should investigate that.

cc: @luisherranz

package.json
```diff
@@ -42,6 +42,7 @@
 	},
 	"dependencies": {
 		"hpq": "^1.3.0",
-		"preact": "^10.10.6"
+		"preact": "^10.10.6",
+		"preact-markup": "^2.1.1"
```
luisherranz: This is not needed anymore, is it?

src/blocks/interactive-child/register-view.js
```diff
@@ -1,3 +1,5 @@
+import 'preact/debug';
```
luisherranz: Can we remove this now?

src/blocks/interactive-parent/register-view.js
```diff
@@ -1,3 +1,5 @@
+import 'preact/debug';
```
luisherranz: Same here.

webpack.config.js
```diff
@@ -1,18 +1,11 @@
 const defaultConfig = require('@wordpress/scripts/config/webpack.config');
 
+delete defaultConfig.optimization.minimizer;
```
luisherranz: You can remove this now.

src/gutenberg-packages/hydration.js
```diff
@@ -1,3 +1,16 @@
...
 const blockViews = createGlobal('blockViews', new Map());
+
+const components = Object.fromEntries(
+	[...blockViews.entries()].map(([k, v]) => [k, v.Component])
+);
+
+visitor.map = components;
+
+const dom = document.querySelector('.wp-site-blocks');
+const vdom = toVdom(dom, visitor, createElement, {}).props.children;
```
luisherranz: Why do you need `.props.children` here?

src/gutenberg-packages/to-vdom.js
```diff
@@ -0,0 +1,68 @@
...
+			if ((!text || text === ' ') && arr.length > 1 && isFirstOrLast)
+				return null;
+		}
+		return text;
+	}
+	if (n.nodeType !== 1) return null;
+	let nodeName = String(n.nodeName).toLowerCase();
+
+	// Do not allow script tags unless explicitly specified
+	if (nodeName === 'script' && !walk.options.allowScripts) return null;
```
luisherranz: Same here. I'm not sure what we should do with scripts, but it shouldn't be optional.

src/gutenberg-packages/to-vdom.js
```diff
@@ -0,0 +1,68 @@
...
+				);
+			}
+			// skip leading/trailing whitespace
+			if ((!text || text === ' ') && arr.length > 1 && isFirstOrLast)
+				return null;
+		}
+		return text;
+	}
+	if (n.nodeType !== 1) return null;
+	let nodeName = String(n.nodeName).toLowerCase();
```
luisherranz, line 10: Why is there a possibility that a node name is not in lowercase?

src/gutenberg-packages/to-vdom.js
```diff
@@ -0,0 +1,68 @@
...
+export default function toVdom(node, visitor, h, options) {
+	walk.visitor = visitor;
+	walk.h = h;
+	walk.options = options || EMPTY_OBJ;
+	return walk(node);
+}
+
+function walk(n, index, arr) {
+	if (n.nodeType === 3) {
+		let text = 'textContent' in n ? n.textContent : n.nodeValue || '';
```
luisherranz: Interesting. What does this thing do?

src/gutenberg-packages/to-vdom.js
```diff
@@ -0,0 +1,68 @@
...
+			} else {
+				text = text.replace(
+					/(^[\s\n]+|[\s\n]+$)/g,
+					walk.options.trim === 'all' || isFirstOrLast ? '' : ' '
+				);
+			}
+			// skip leading/trailing whitespace
+			if ((!text || text === ' ') && arr.length > 1 && isFirstOrLast)
+				return null;
+		}
```
luisherranz: We need to review this carefully in a later PR. I'm not sure if we need trim or not. And if we do, I doubt it should be optional.

src/gutenberg-packages/to-vdom.js
```diff
@@ -0,0 +1,68 @@
...
+	return out;
+}
+
+function getProps(attrs) {
+	let len = attrs && attrs.length;
+	if (!len) return null;
+	let props = {};
+	for (let i = 0; i < len; i++) {
+		let { name, value } = attrs[i];
+		if (name.substring(0, 2) === 'on' && walk.options.allowEvents) {
```
luisherranz: Same here. We don't need options.

luisherranz (approved):
Looks great, David!

I think it's ready to merge, and we can continue in other PRs. I've asked a few questions, though.

Also, I'm not sure yet if it'd be better to use the `visitor` pattern or Preact `options`. I'm inclined to think Preact `options` are safer.

src/gutenberg-packages/hydration.js
DAreRodz: Because you run `hydrate` on the parent element (`wp-site-blocks`), but you pass the vdom of the content (`wp-site-blocks` works as a wrapper).

We can change this with a better approach in the future. 🙂

src/gutenberg-packages/to-vdom.js
DAreRodz: If the node is an [Element](https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeName#:~:text=of%20DocumentType.name-,Element,tag%20if%20an%20XML%20element%20(like%20a%20SVG%20or%20MATHML%20element).,-ProcessingInstruction), it _always_ return the uppercase name (e.g. `DIV` for `<div>`). 😅

src/gutenberg-packages/to-vdom.js
DAreRodz: Just copy-pasted from `preact-markup`. I've refactored that part.

src/gutenberg-packages/to-vdom.js
DAreRodz: That was causing hydration problems; I've removed all that code.

src/gutenberg-packages/to-vdom.js
DAreRodz: Options removed!

src/gutenberg-packages/to-vdom.js
DAreRodz: Options removed. 🧹

block-hydration-experiments.php
```diff
@@ -116,7 +118,12 @@ function bhe_block_wrapper($block_content, $block, $instance)
 		$template_wrapper,
 		sprintf($block_wrapper, $block_content . $empty_template)
 	);
-	return sprintf($block_wrapper, $block_content);
+
+	// The block content comes between two line breaks that seem to be included during block
+	// serialization, corresponding to those between the block markup and the block content.
+	//
+	// They need to be removed here; otherwise, the preact hydration fails.
+	return sprintf($block_wrapper, substr($block_content, 1, -1));
```
DAreRodz, line 126: I've left a note here explaining the fix for the Preact hydration.

src/gutenberg-packages/hydration.js
luisherranz: Oh, I see. I'm using [createRootFragment](https://gist.github.com/developit/f4c67a2ede71dc2fab7f357f39cff28c) for that 🙂

src/gutenberg-packages/to-vdom.js
luisherranz: Ok. I was using `node.localName` 🙂

block-hydration-experiments.php
luisherranz: This should be fixed in Gutenberg itself.

PR #68, opened 2022-09-15
⚛️ Add directives to handle `providesContext` and `usesContext`

### LINKED ISSUE #67: ⚛️ Share block context between blocks
Right now, blocks have two properties called `data-wp-block-provides-block-context` and `data-wp-block-uses-block-context`, containing the information of those attributes that are exposed and consumed.

The idea is to use the [Option Hooks API](https://preactjs.com/guide/v10/options/) from Preact to handle those properties as directives.

DAreRodz:
Closes #67

I've refactored `toVdom` significantly and added support for directives like in https://github.com/WordPress/gutenberg/pull/44034. There are still some code I haven't moved to directives, like sourced attributes generation and inner blocks location (not sure if it is possible).

Also, I created three directives:

- One that handles `type` and renders the block component (if it exists).
- Another one that handles `providesContext` and renders a wrapper around block children.
- And the last one, which handles `usesContext`, that injects a `context` property to a block component using a HOC.

Directives are placed in `src/directives`.


src/gutenberg-packages/to-vdom.js
```diff
@@ -1,41 +1,113 @@
...
+	if (type === 'wp-block' && innerBlocksFound) {
+		wpBlock.innerBlocks = innerBlocksFound;
+		innerBlocksFound = null;
+
+		// Set wpBlock prop again, just in case it's missing.
+		props.wpBlock = wpBlock;
+	}
+
+	// Create vNode. Note that all `wpBlock` props should exist now to make directives work.
+	const vNode = h(type, props, children);
```
DAreRodz, line 59: Here I've assumed that `options.vnode` would run every time `h()` is used. Maybe I'm wrong here. 🤔

luisherranz: You're not. [That's the way it works](https://github.com/preactjs/preact/blob/master/src/create-element.js#L79).

luisherranz (approved):
At first glance, it looks great to me. I'm glad the current implementation of the directives is flexible enough to inject atomic functionality into the vdom. However, I was hoping we could delegate more to the wrapper component and less to the vdom internals.

But we can keep playing with this API in other PRs.

DAreRodz:
> However, I was hoping we could delegate more to the wrapper component and less to the vdom internals.

I left three things inside `toVdom` internals:
1. Sourced attributes computation ― it needs access to the original DOM, at least with the current implementation.
2. Block props processing (`class` to `className`, `style` string to object, etc.) ― I guess we can move that inside the wrapper component without much problem.
3. Inner blocks location ― this can also be moved to the wrapper component, although I found it convenient to do that during `toVdom()` execution. We can give this a thought in the future.

Thanks for your review @luisherranz. I'm merging this PR. 😄

PR #71, opened 2022-09-19
⚛️ Fix hydration when there are scripts in the top of the body

luisherranz:
@darerodz and I just discovered that hydration could fail if there are `<script>` tags at the top of the `<body>`, so instead of destroying them we can just return an empty `<script>`.


luisherranz:
Also fixed in the Query Loop block experiment:

- https://github.com/WordPress/gutenberg/pull/44034/commits/3104b61bddd46e8f1ae5f4451dec202332b3fb17

DAreRodz (approved):
Approved! 👍

PR #72, opened 2022-09-21
⚛️ Get rid of the `<wp-block>` wrapper

DAreRodz:
This PR removes the `<wp-block>` wrapper and, instead, appends all wp block attributes to the element's attributes where `blockProps` are rendered.

The technique is a bit hacky because:
1. there is no way to render those attributes during `save`, as most of the required information is not available
2. attributes are appended to the block element modifying the saved content with `grep_replace()`, using a RegExp that could not be reliable enough.

### DIFF
--- a/block-hydration-experiments.php
+++ b/block-hydration-experiments.php
@@ -119,11 +119,20 @@ function bhe_block_wrapper($block_content, $block, $instance)
 
 	// Append all wp block attributes after the class attribute containing the block class name.
 	// Elements with that class name are supposed to be those with the wrapper attributes.
-	// Note that $class_pattern could not cover some edge cases.
 	//
 	// We could use `WP_HTML_Walker` (see https://github.com/WordPress/gutenberg/pull/42485) in the
 	// future if that PR is finally merged.
-	$class_pattern     = '/class="[\w\s-]*wp-block-[\w-]+[\w\s-]*"/';
+
+	// This replace is similar to the one used in Gutenberg (see
+	// https://github.com/WordPress/gutenberg/blob/1582c723f31ce0f728cd4dcc1af37821342eeaaa/packages/blocks/src/api/serializer.js#L41-L44).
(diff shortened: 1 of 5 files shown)


block-hydration-experiments.php
```diff
@@ -92,38 +92,42 @@ function bhe_block_wrapper($block_content, $block, $instance)
...
+
+	// Append all wp block attributes after the class attribute containing the block class name.
+	// Elements with that class name are supposed to be those with the wrapper attributes.
+	// Note that $class_pattern could not cover some edge cases.
+	//
+	// We could use `WP_HTML_Walker` (see https://github.com/WordPress/gutenberg/pull/42485) in the
+	// future if that PR is finally merged.
+	$class_pattern     = '/class="[\w\s-]*wp-block-[\w-]+[\w\s-]*"/';
+	$class_replacement = '$0 ' . $block_wrapper_attributes;
+	$block_content     = preg_replace( $class_pattern, $class_replacement, $block_content, 1 );
```
DAreRodz: Another thing I've thought of is to add an empty attribute during block saving, something like `wp-block-wrapper-attributes`, and replace it here with the `$block_wrapper_attributes`.

That could be more reliable than using the `$class_pattern` above.

luisherranz: Can we use the class name inside the regular expression?

DAreRodz: I think so; I'll try to do it.

luisherranz (approved):
Looks great to me. Looking forward to the day we have the `WP_HTML_Walker` available for these HTML injections 🙂

It may be worth noting that this would only work if we enforce blocks to have a single node wrapper. Maybe we need to add that to the requirements in the `README`.

luisherranz:
> Maybe we need to add that to the requirements in the README.

Added in https://github.com/WordPress/block-hydration-experiments/pull/72/commits/110dcbf5287dd46835fa2555460140b83f5dbe0d.

Issue #59, opened 2022-08-25
Make sure inner blocks are sent to the frontend even if the Save/View component doesn't render them initially


DAreRodz:
Right now, when inner blocks are hidden in the editor, they are not serialized and stored in the WP database, so that information is lost. Thus, the block cannot interactively show inner blocks later in the browser.

There were some previous attempts to solve this by adding a `<template>` tag.

- https://github.com/WordPress/block-hydration-experiments/pull/8

However, we removed the `<template>` tag on #28 (see https://github.com/WordPress/block-hydration-experiments/pull/28/commits/fc6fa92f96f0e60f093605449c45d4157927b3c1).

Related to https://github.com/WordPress/block-hydration-experiments/issues/29.

luisherranz:
Should we try to do this in the Gutenberg repo instead? Maybe in the [serializer](https://github.com/WordPress/gutenberg/blob/trunk/packages/element/src/serialize.js)? I'm not sure if it makes sense to do it in _user land_ when it won't be necessary...

luisherranz:
For anyone interested in taking a look at this, I think that what needs to happen is:

![image](https://user-images.githubusercontent.com/3305402/186853188-6ecf0057-593b-4de3-b387-749cbf33b330.png)

[Excalidraw](https://excalidraw.com/#json=QscBFTeWkWB3pqg-H0U7Y,Lwr7Db16H6V9tOUojlqBoQ)

Non-interactive blocks don't need this because those "hidden" inner blocks won't ever be shown, and therefore they don't need to be sent to the client.

luisherranz:
Closed as we're not actively working on this experiment anymore and this works fine in the Directives Hydration.

PR #74, opened 2022-09-22
⚛️ Refactor directives and make them work inside client components

### LINKED ISSUE #73: ⚛️ Directives don't work if they are used inside of a component
Because we're extracting the `wp-` attributes only on the server-side rendered HTML, if a component contains a directive, it won't work:

```js
const View = ({ ... }) => (
  <div {...}>
    <span wp-some-directive>...</span>
  </div>
);
```

I think we need to move everything to `options.vnode`. I'll open a PR to investigate.

luisherranz:
Fixes #73.

For now, I've just added a simple directive to reproduce the problem. I'll continue working on the solution.


luisherranz:
I don't like having to traverse all the `props` again in `options.vnode` to search for the ones that start with `wp-`, but I don't see another way around it. Other ideas are welcomed 🙂

luisherranz:
Well, there are actually options to avoid the double traverse:

- Marking the static vnodes to bypass the traverse in the options hooks.
- Attach the `attributes` to the `vnode` and move all the traverse to the options hooks.

I think I like the first option better because with the second the static vnodes won't have props until our options hook runs.

luisherranz:
For future reference, a couple of perf tests:

- [From attributes array to props object](https://perf.link/#eyJpZCI6InZuNjc1czFoMGZhIiwidGl0bGUiOiJNb3ZlIGF0dHJpYnV0ZXMgdG8gcHJvcHMiLCJiZWZvcmUiOiJjb25zdCBhdHRyaWJ1dGVzID0gWy4uLkFycmF5KDEwKS5rZXlzKCldLm1hcCgoKSA9PiAoe1xuXHRuYW1lOiBNYXRoLnJhbmRvbSgpLFxuXHR2YWx1ZTogTWF0aC5yYW5kb20oKSxcbn0pKTtcbmNvbnN0IHByb3BzID0ge307IiwidGVzdHMiOlt7Im5hbWUiOiJmb3IiLCJjb2RlIjoiZm9yIChsZXQgaSA9IDA7IGkgPCBhdHRyaWJ1dGVzLmxlbmd0aDsgaSsrKSB7XG5cdHByb3BzW2F0dHJpYnV0ZXNbaV0ubmFtZV0gPSBhdHRyaWJ1dGVzW2ldLnZhbHVlO1xufSIsInJ1bnMiOls1MzcwMDAsMzMzMDAwLDE2NDAwMCw0OTQwMDAsNTI2MDAwLDQ4MzAwMCwzNDYwMDAsMTI5MDAwLDI0ODAwMCwyODAwMCw3MTYwMDAsNTc2MDAwLDIwOTAwMCw1NDIwMDAsNjg4MDAwLDcwODAwMCw2NTkwMDAsNDMwMDAwLDk1MDAwLDYxMTAwMCw3MDcwMDAsNTE3MDAwLDkyMDAwLDUwMDAsNjY4MDAwLDQzNTAwMCwyNTYwMDAsNTAxMDAwLDY1MDAwLDIwMDAsNTMwMDAsMTc2MDAwLDcwNTAwMCw1MzQwMDAsMTAwMCw2NzQwMDAsNjA1MDAwLDU3MjAwMCw1NTAwMCw1ODcwMDAsNDk2MDAwLDI1MDAwLDYwNTAwMCwxMDAwLDY4MDAwMCw1ODYwMDAsNzE2MDAwLDM5MDAwMCw3MzcwMDAsNTgwMDAwLDUzNjAwMCw1MzcwMDAsNzkwMDAsNjUzMDAwLDQ5MzAwMCwxMzgwMDAsNzM3MDAwLDYxNDAwMCw3MzAwMCw2NDgwMDAsNjU0MDAwLDU0OTAwMCw3MjAwMDAsNDM3MDAwLDQ5MTAwMCw0NjIwMDAsNDQ0MDAwLDI0OTAwMCw2NTMwMDAsNjI3MDAwLDYzMjAwMCwxNjQwMDAsMTkwMDAsMTIxMDAwLDcxMDAwMCw2NDEwMDAsNTYzMDAwLDEwMDAsMzIyMDAwLDEwOTAwMCwzODAwMCw2MjcwMDAsNTM3MDAwLDUzNzAwMCw5OTAwMCw0NTUwMDAsNjAwMCwzMzAwMCwyNTAwMCwyMDAwLDIwMDAsMTMwMDAsMTUxMDAwLDg3MDAwLDY2MDAwLDUzNzAwMCw0ODgwMDAsNzIxMDAwLDE3MDAwLDcwMjAwMF0sIm9wcyI6Mzg3NjcwfSx7Im5hbWUiOiJmb3IgZWFjaCIsImNvZGUiOiJhdHRyaWJ1dGVzLmZvckVhY2goKGF0dHIpID0%2BIHtcblx0cHJvcHNbYXR0ci5uYW1lXSA9IGF0dHIudmFsdWU7XG59KTsiLCJydW5zIjpbNDc3MDAwLDQyMDAwLDE3MDAwLDMyNzAwMCw4NTIwMDAsNDk4MDAwLDE2ODAwMCwyNTAwMCw4NjAwMCw2MTIwMDAsNzI0MDAwLDI4NTAwMCwxMzYwMDAsMTkwMDAwLDg3MzAwMCw5MjcwMDAsNTM5MDAwLDQwMjAwMCw4NTUwMDAsNTMyMDAwLDUxNzAwMCwyMjcwMDAsNzIzMDAwLDQzNTAwMCw1ODcwMDAsNDQzMDAwLDI2NTAwMCw0MDAwMCw3MDAwMDAsNTc1MDAwLDE2MzAwMCw5MDkwMDAsNTAyMDAwLDQ2MTAwMCw2MTUwMDAsNDcxMDAwLDM5MDAwMCwzMzcwMDAsNTA0MDAwLDM5OTAwMCwzMTcwMDAsNjI1MDAwLDg1MzAwMCw2NDgwMDAsNDc5MDAwLDU1MTAwMCwyMTUwMDAsODgyMDAwLDU4MzAwMCwzODkwMDAsMzk3MDAwLDI1NTAwMCw5MTEwMDAsNzY3MDAwLDM2ODAwMCw5MDIwMDAsNzkxMDAwLDEwOTAwMCw2MzMwMDAsNjEwMDAwLDMzNzAwMCwzMjAwMDAsNzczMDAwLDIyNDAwMCwxMTEwMDAsODkwMDAsMzc3MDAwLDk5MTAwMCw4MjMwMDAsNDU0MDAwLDEyNDAwMCw2NTYwMDAsNDI1MDAwLDc2NjAwMCwzODAwMDAsMjQ3MDAwLDM0NDAwMCwyMDAwMCwxMDYwMDAsNzQ2MDAwLDU1OTAwMCw3MDUwMDAsMzg4MDAwLDI1MzAwMCw3MTkwMDAsNzMwMDAwLDUxMjAwMCw2ODgwMDAsNTQ4MDAwLDc2NzAwMCw3NDAwMDAsNzQyMDAwLDE2MDAwMCw4MjcwMDAsNTIyMDAwLDQ3MTAwMCwyNjkwMDAsMTAwMCw0ODcwMDAsNTU0MDAwXSwib3BzIjo0ODA3MDB9LHsibmFtZSI6ImZvciBvZiIsImNvZGUiOiJmb3IgKGNvbnN0IGF0dHIgb2YgYXR0cmlidXRlcykge1xuXHRwcm9wc1thdHRyLm5hbWVdID0gYXR0ci52YWx1ZTtcbn1cbiIsInJ1bnMiOlsyNTYwMDAsMjcwMDAsMzgyMDAwLDk0MDAwLDQ3MTAwMCw2MTAwMCw1NDUwMDAsNDY2MDAwLDY2NTAwMCwxODYwMDAsMjgyMDAwLDY4ODAwMCwxMDAwLDUxNDAwMCw0NDgwMDAsNzY4MDAwLDIyNjAwMCw3NDQwMDAsNTM1MDAwLDMzNDAwMCw2NDAwMCw2NDkwMDAsNDQxMDAwLDczOTAwMCwzNTIwMDAsNjk1MDAwLDIwOTAwMCw2NTkwMDAsMzY0MDAwLDE1NzAwMCw4MDAwMCw0NDgwMDAsMTQzMDAwLDU4MDAwMCw1NjAwMCwxNzMwMDAsMjkwMDAsNjc0MDAwLDE5MjAwMCwyMDAwLDYwOTAwMCw0MDAwLDExMDAwMCwxNTQwMDAsMTc3MDAwLDEyNDAwMCw5NzAwMCw3NDQwMDAsMTU2MDAwLDM0NDAwMCw5MjAwMCw3MDUwMDAsNDQ4MDAwLDE4NTAwMCw2OTkwMDAsMzM3MDAwLDM5MjAwMCw0OTUwMDAsMzA4MDAwLDMwNDAwMCwzMDAwLDY5OTAwMCw0NDgwMDAsMTU3MDAwLDQzODAwMCwxMjAwMCw4MjAwMCwzNjAwMDAsNTMxMDAwLDI3NzAwMCw2NzEwMDAsMjkzMDAwLDc2OTAwMCw0NDIwMDAsMTc2MDAwLDgzMDAwLDU0MTAwMCwxNTAwMCw1MTgwMDAsMTczMDAwLDM3ODAwMCwzOTcwMDAsNzAwMCw1ODQwMDAsNDMzMDAwLDQyMjAwMCwxNTMwMDAsMTI0MDAwLDQ0ODAwMCwyMTMwMDAsMzA1MDAwLDI3MjAwMCw2MTUwMDAsMjU3MDAwLDMxMjAwMCwxMTAwMCw3NDUwMDAsNDMzMDAwLDE4ODAwMCw0MzMwMDBdLCJvcHMiOjM0MzAxMH1dLCJ1cGRhdGVkIjoiMjAyMi0wOS0yMlQxMzoxNzowOC45NjJaIn0%3D)
- [Check if any prop keys start with `wp-`](https://perf.link/#eyJpZCI6Ing0aTEzZGh6MmVyIiwidGl0bGUiOiJGaW5kaW5nIGtleXMgdGhhdCBzdGFydCB3aXRoIFwid3AtXCIiLCJiZWZvcmUiOiJjb25zdCBvYmogPSB7fTtcbmZvciAobGV0IGkgPSAwOyBpIDwgMTA7IGkrKykge1xuXHRvYmpbTWF0aC5yYW5kb20oKV0gPSBNYXRoLnJhbmRvbSgpO1xufVxuY29uc3Qga2V5cyA9IE9iamVjdC5rZXlzKG9iaik7IiwidGVzdHMiOlt7Im5hbWUiOiJmb3IiLCJjb2RlIjoiZm9yIChsZXQgaSA9IDA7IGkgPCBrZXlzLmxlbmd0aDsgaSsrKSB7XG4gIGtleXNbaV0uc3RhcnRzV2l0aCgnd3AtJyk7XG59IiwicnVucyI6WzUwNzAwMCw1ODAwMDAsNzI5MDAwLDEzMjAwMCwxNTQ1MDAwLDIwOTUwMDAsMTM3MTAwMCwxMjAwMCwxMTUyMDAwLDIyNTUwMDAsMTM3MjAwMCwxMjc0MDAwLDI5MTAwMCwyMTI5MDAwLDUzMDAwMCwxNzk0MDAwLDEyMDgwMDAsMjAwMCwxNDcyMDAwLDE1MTAwMDAsNTExMDAwLDc2MzAwMCwxNDQ4MDAwLDE5ODEwMDAsMTA2NjAwMCw3MjAwMDAsMTE1MTAwMCw5NzIwMDAsMjAwMCwxNTU0MDAwLDc2MDAwLDE2NDgwMDAsMTQ5MjAwMCwxODY3MDAwLDEyOTUwMDAsMjEwNzAwMCwxNTQwMDAwLDExNjkwMDAsMTg1OTAwMCw5MDEwMDAsMTA3MDAwLDgzOTAwMCwxNjg3MDAwLDk3MjAwMCwyMDIzMDAwLDEzMjkwMDAsMjI4MDAwMCw5NzMwMDAsMTAxMDAwLDcyOTAwMCwyMDE2MDAwLDEwNTgwMDAsMjAwNDAwMCwxNjA5MDAwLDk3MjAwMCw5NDAwMCwxNDIwMDAwLDIxMzkwMDAsMTA1ODAwMCw1OTAwMCwyMDU2MDAwLDE2ODkwMDAsMTQ4NTAwMCwxNzEzMDAwLDk0NTAwMCwxNzk4MDAwLDEzNjAwMDAsODQ2MDAwLDE2OTcwMDAsNDExMDAwLDEyNzAwMCwxMjI0MDAwLDIwMzcwMDAsMTM1OTAwMCw0NzcwMDAsMTkwMDAsMTkwMDAsMjIyNDAwMCw3MDcwMDAsMjI2NzAwMCw2ODEwMDAsMjAyMjAwMCw2NDYwMDAsMTg1NTAwMCwxMTgwMDAsMjAxNDAwMCw3MzAwMCwxOTM1MDAwLDExMDgwMDAsMTU0MTAwMCwxMTg0MDAwLDE2MTkwMDAsMTg0NjAwMCwxMDk3MDAwLDE3MjQwMDAsOTE2MDAwLDIzMTAwMDAsMTA0MjAwMCwxNzM2MDAwLDEwNzgwMDBdLCJvcHMiOjEyMDU1NjB9LHsibmFtZSI6ImZvciBvZiIsImNvZGUiOiJmb3IgKGNvbnN0IGtleSBvZiBrZXlzKSB7XG5cdGtleS5zdGFydHNXaXRoKCd3cC0nKTtcbn1cbiIsInJ1bnMiOls4MjEwMDAsMjMwMDAsMzc2MDAwLDcyMDAwLDUwNjAwMCwxNjIwMDAwLDUxMTAwMCwxNzUyMDAwLDcxMDAwMCwxMjY1MDAwLDExNTAwMCw1NTcwMDAsMTgxMDAwLDExMzQwMDAsNzgwMDAsODIxMDAwLDMyNzAwMCw0OTQwMDAsMzExMDAwLDg3ODAwMCwyNzYwMDAsNTMyMDAwLDM3MTAwMCw0OTIwMDAsMzAwMDAwLDM1MjAwMCw2MjAwMDAsOTEwMDAsMTAwOTAwMCw2MjMwMDAsMTI2NTAwMCw3MTYwMDAsMzI1MDAwLDkyNjAwMCwyMTEwMDAsMTI2MDAwMCwxMDA5MDAwLDUwNjAwMCw2NDAwMDAsMjg5MDAwLDEyNDkwMDAsNTk0MDAwLDQyMDAwMCw0NTYwMDAsOTU1MDAwLDQ5NjAwMCw5NDcwMDAsMjMzMDAwLDk1MDAwMCwxNTQwMDAwLDExMTUwMDAsMzY5MDAwLDEwMTAwMDAsMzgwMDAwLDMzODAwMCwxMzAxMDAwLDU4NTAwMCwxMTY3MDAwLDE4MjAwMCwxMDA5MDAwLDU4MzAwMCw1MDYwMDAsNDkwMDAwLDc2NzAwMCwxODE2MDAwLDk5NjAwMCw1MDEwMDAsMTQ1OTAwMCw5NTkwMDAsMTcwMDAsMTI2ODAwMCw0NjAwMDAsMTE4MjAwMCw1MDYwMDAsMjcxMDAwLDY1MzAwMCw5ODgwMDAsNzEzMDAwLDQ2MTAwMCwxMzE1MDAwLDEwODIwMDAsMTAwOTAwMCwxNjQyMDAwLDc1NDAwMCwxNTEyMDAwLDk2MzAwMCw2NjUwMDAsMTMzMjAwMCw3NTYwMDAsOTYzMDAwLDIxMjAwMCw2OTQwMDAsOTA1MDAwLDUwNjAwMCw4NTkwMDAsMzU2MDAwLDEyOTUwMDAsMTA2NzAwMCwxNzM3MDAwLDQzMzAwMF0sIm9wcyI6NzQzMTQwfSx7Im5hbWUiOiJmb3JFYWNoIiwiY29kZSI6ImtleXMuZm9yRWFjaCgoa2V5KSA9PiB7XG5cdGtleS5zdGFydHNXaXRoKCd3cC0nKTtcbn0pOyIsInJ1bnMiOls3MDEwMDAsMTI0NzAwMCw1MzYwMDAsMTc4NTAwMCw1NDMwMDAsMjEzMDAwMCwzODcwMDAsMTYyODAwMCwxODA0MDAwLDY2MDAwMCwxODI0MDAwLDU2NTAwMCwxOTMzMDAwLDg2MzAwMCwxODAzMDAwLDgwMDAwMCwxNTU2MDAwLDc5NDAwMCwyOTUwMDAsNTA1MDAwLDE4NzEwMDAsMTAzMTAwMCwxODA0MDAwLDM5MjAwMCw2NzAwMCwyMTE5MDAwLDcwNTAwMCwxNzQ4MDAwLDEwNTEwMDAsMjIxOTAwMCwxODgwMDAsNDYxMDAwLDkyMDAwLDExNjAwMCwxNzE5MDAwLDcxNTAwMCwyMTMyMDAwLDE2MDgwMDAsMzQ1MDAwLDIxMTQwMDAsMTI4ODAwMCwxNTIwMDAsMzk3MDAwLDE5NTgwMDAsNTEyMDAwLDE4MzkwMDAsNzg5MDAwLDExMzYwMDAsMTEzMjAwMCwxMjgzMDAwLDE4MDQwMDAsMTkwMDAsMTI5NDAwMCwxMTYwMDAsMTgwNDAwMCwzNTAwMDAsMjEzNzAwMCwxNDE2MDAwLDIwNTcwMDAsMTY0OTAwMCwzOTUwMDAsMzMzMDAwLDIxMzkwMDAsNDUwMDAsMTk0ODAwMCw1NTEwMDAsMTQyMDAwLDE2NDEwMDAsNTgyMDAwLDEyMDUwMDAsMTM0MDAwMCwzMzEwMDAsMTc0NDAwMCwyMjMwMDAsMTYxNTAwMCwxODUwMDAsNjUwMDAwLDgwODAwMCwyMDA2MDAwLDEwMjcwMDAsNjMyMDAwLDEwMTUwMDAsMTc3OTAwMCwyODQwMDAsNzY3MDAwLDU0NDAwMCwyMDQ1MDAwLDgxNzAwMCwxODU3MDAwLDMxMjAwMCwxMDcyMDAwLDM5MjAwMCw4NTQwMDAsNDk4MDAwLDgxNjAwMCwzNzgwMDAsMTI2ODAwMCw0OTEwMDAsMTg4MjAwMCwxMDAwXSwib3BzIjoxMDQ2MDIwfV0sInVwZGF0ZWQiOiIyMDIyLTA5LTIyVDEzOjE4OjEwLjI1N1oifQ%3D%3D)

I tested it in Chrome desktop and mobile, Firefox desktop and mobile and Safari desktop. 
- In Chrome, `forEach` is slightly faster but not for much.
- In Safari, `forEach` is quite slow. `for` and `for of` are similar.
- In Firefox, `for` is faster.

`for` seems to be the best option in both cases.

luisherranz:
I think everything is finished here. Ready for review.

<a href="https://www.loom.com/share/4f9809e14edc45b49024014a92086862">
    
  </a>

https://www.loom.com/share/4f9809e14edc45b49024014a92086862

luisherranz:
Oh, something is going on with the block context. It doesn't work after hiding+showing.  I'll take a look.

https://user-images.githubusercontent.com/3305402/192328406-39b31d09-d7c0-4745-a59c-460ef033cf62.mp4

DAreRodz:
> Oh, something is going on with the block context. It doesn't work after hiding+showing. I'll take a look.

~Hey @luisherranz, I tried to reproduce the issue you mentioned, but after hiding and showing the blocks, the context still works. The blocks were arranged similarly. 😕~

I also tested it in `main-directives-hydration`, just in case, and it also worked.

EDIT: my bad, I was in the wrong commit (https://github.com/WordPress/block-hydration-experiments/pull/74/commits/9057066149ff93a3a3410e5651e459bdbb6c44c5). 🤦 In the latest commit, the issue is reproducible.

DAreRodz (commented):
@luisherranz, the code seems great. 👍

Regarding the issue, I don't know where it could be TBH. 😕 Maybe it is related to the attributes being processed now inside the `wpType` directive, or re-processing the `wp` directives on non-static nodes...

No idea. 🙃

DAreRodz:
I've taken another quick look at the block context issue, and I think it should be related to the way sourced attributes are [computed](https://github.com/WordPress/block-hydration-experiments/blob/5104a18e7fe782b73fc7fb4d07b843f9473af09a/src/directives/wp-block.js#L48-L55) after the refactoring (inside the `blockType` directive).

It is like the `title` attribute disappears when the node is mounted again.

luisherranz:
Thanks, David. I'll take a look later today 🙂

luisherranz:
> It is like the title attribute disappears when the node is mounted again.

It disappears because it's a sourced attribute and needs to look at the DOM to get the value. At the beginning of the application, we have a DOM and create a vDOM. So we go DOM-> vDOM and the sourced attribute values can be found. But when Preact components add nodes, we go in the other direction, vDOM-> DOM. So there's no DOM to search for the sourced attribute values.

I'm not sure why or how this was working previously. I'll take a look.

luisherranz:
> I'm not sure why or how this was working previously. I'll take a look.

Ok, it makes sense now. You were processing the sourced attributes in the `toVdom` function, which means they will exist (and persist) in the vnode children (`vnode.props.children`) that then the component hides/shows using `{show && children}`. I'm processing them in the `options.vnode` function, which means they are processed on the fly each time a `vnode` is created, and therefore it needs to process them again when there's no DOM yet.

I'll think about how we can do this without hardcoding that behavior into the `toVdom` function.

luisherranz:
I've finally exposed the `vnode` and I'm using it a storage medium for the source attributes. Of course I don't love it, but it's the only thing that is stable and at least it's not hardcoded. We can think about better alternatives in the future.

I've also added support for returning components.

This should be ready for review! 

<a href="https://www.loom.com/share/e71a9867720542eaa5812d70741ae4fd">
    
  </a>

https://www.loom.com/share/e71a9867720542eaa5812d70741ae4fd

cc: @darerodz

luisherranz:
EDIT: The video is finally loading fine.

src/directives/wp-block.js
```diff
@@ -45,24 +46,38 @@ directive('blockType', (props) => {
...
+		vnode._blockAttributes = blockAttributes || {};
+		useMemo(() => {
+			for (const attr in blockSourcedAttributes) {
+				vnode._blockAttributes[attr] = matcherFromSource(
+					blockSourcedAttributes[attr]
+				)(initialRef);
+			}
+		}, [blockAttributes, blockSourcedAttributes]);
+	}
+	props.attributes = vnode._blockAttributes;
```
DAreRodz: I was intrigued by that `useMemo`, and I think we may be able to get rid of it.

We are using `vnode` as a container to store`_blockAttributes`. That `vnode` —the original one returned by `toVdom()`— is not recreated and always exists. So, once you set the `_blockAttributes` prop, I guess that code won't run again, am I right? 🤔

Apart from that, `useMemo` is used conditionally, and that's not supposed to be the way. 😅

DAreRodz (approved):
Great refactoring, thanks @luisherranz!! 👏 

PS: I left a comment below.

src/directives/wp-block.js
luisherranz: You're totally right. Fixed.

I've also added a comment to explain why we're doing such a weird thing.

I'm also curious how sourced attributes will play out with the initially hidden blocks (the ones that [will use `<template>` in the future](https://github.com/WordPress/block-hydration-experiments/issues/59)).

luisherranz:
Thanks, David!! 🎉

PR #77, opened 2022-09-27
⚛️ Use a `wp-inner-block` attribute for each inner block

### LINKED ISSUE #76: ⚛️ Remove the `wp-inner-blocks` wrapper
We need to get rid of the `<wp-inner-blocks>` element to avoid CSS issues like https://github.com/WordPress/block-hydration-experiments/issues/50.

As @luisherranz proposed in https://github.com/WordPress/block-hydration-experiments/issues/50#issuecomment-1235334619, we can append the `wp-inner-block` attribute to those blocks that are the children of an interactive block.

```html
<div class="wp-block-parent">
  <div>Some content</div>
  <div class="wp-block-child-1" wp-inner-block>Some child content</div>
  <div class="wp-block-child-2" wp-inner-block>More child content</div>
</div>
```

Note that we need a new attribute because:
- using a regexp to find nodes with their class starting with `wp-block-"` could be not reliable enough
- using the `wp-block-type` attribute would work only if children are interactive blocks as well

DAreRodz:
Will fix #76

### DIFF
--- a/block-hydration-experiments.php
+++ b/block-hydration-experiments.php
@@ -54,7 +54,7 @@ add_filter('render_block_bhe/interactive-parent', function ($content) {
 function bhe_block_wrapper($block_content, $block, $instance)
 {
 	// Append the `wp-inner-block` attribute for inner blocks of interactive blocks.
-	if ( isset($instance->parsed_block['isInnerBlock']) ) {
+	if (isset($instance->parsed_block['isInnerBlock'])) {
 		$block_content = bhe_append_block_attributes(
 			$instance->name,
 			$block_content,
@@ -102,23 +102,22 @@ function bhe_block_wrapper($block_content, $block, $instance)
 	WP_Block_Supports::$block_to_render = $previous_block_to_render;
 
 	// Generate all required wrapper attributes.
-	$block_wrapper_attributes =
-		sprintf(
-			'data-wp-block-type="%1$s" ' .
+	$block_wrapper_attributes = sprintf(
+		'data-wp-block-type="%1$s" ' .
 			'data-wp-block-uses-block-context="%2$s" ' .
 			'data-wp-block-provides-block-context="%3$s" ' .
 			'data-wp-block-attributes="%4$s" ' .
 			'data-wp-block-sourced-attributes="%5$s" ' .
 			'data-wp-block-props="%6$s" ' .
 			'data-wp-block-hydration="%7$s"',
-			esc_attr($block['blockName']),
-			esc_attr(json_encode($block_type->uses_context)),
-			esc_attr(json_encode($block_type->provides_context)),
-			esc_attr(json_encode($attributes)),
-			esc_attr(json_encode($sourced_attributes)),
-			esc_attr(json_encode($block_props)),
-			esc_attr($hydration_technique)
-		);
+		esc_attr($block['blockName']),
+		esc_attr(json_encode($block_type->uses_context)),
+		esc_attr(json_encode($block_type->provides_context)),
+		esc_attr(json_encode($attributes)),
+		esc_attr(json_encode($sourced_attributes)),
+		esc_attr(json_encode($block_props)),
+		esc_attr($hydration_technique)
+	);
 
 	// Append block wrapper attributes.
 	$block_content = bhe_append_block_attributes(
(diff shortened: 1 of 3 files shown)


DAreRodz:
We cannot use a regexp that detects the class name for inner blocks, as there could be blocks without it (e.g., Paragraph).

However, we can mark all inner blocks of interactive blocks, adding a callback to the `render_block_data` hook and "polluting" the parsed block data (`$parsed_block`) with a flag that indicates it is an inner block.

https://github.com/WordPress/block-hydration-experiments/blob/fe4994b561c7d82a558edc24cd630611b9ea03f7/block-hydration-experiments.php#L140-L154

Then, we can read that flag during `render_block` and append the `wp-inner-block` attribute to the block wrapper. Again, we cannot use the block class name to detect the wrapper: we should assume it is the first HTML tag or it won't work for those blocks that don't have a class name.

DAreRodz:
There's an issue with inner blocks' hydration. I'm taking a look.

DAreRodz:
> There's an issue with inner blocks' hydration. I'm taking a look.

Some text nodes with white spaces were making hydration fail (again). I fixed it by preserving those nodes that appeared between inner blocks.

src/gutenberg-packages/to-vdom.js
```diff
@@ -46,6 +46,15 @@ export default function toVdom(n) {
 	// Walk child nodes and return vDOM children.
 	const children = [].map.call(n.childNodes, toVdom).filter(exists);
 
+	// Create an array of inner blocks if they are found in children. All nodes in between (e.g.,
+	// line breaks, white spaces, etc.) are preserved in order to prevent hydration failure.
+	const isInnerBlock = ({ props }) => props && 'wp-inner-block' in props;
+	if (children.some(isInnerBlock)) {
```
luisherranz, line 52: Note for the future: this is probably not the best approach because, although `key in obj` is fast, it's doing it multiple times.

If we continue with this approach, we could create a nice set of benchmarks for the `toVdom` function 🙂

src/gutenberg-packages/to-vdom.js
```diff
@@ -46,6 +46,15 @@ export default function toVdom(n) {
...
+		const first = children.findIndex(isInnerBlock);
+		const last = children.findLastIndex(isInnerBlock);
+		innerBlocksFound = children.slice(first, last + 1);
```
luisherranz, line 55: Note for the future: As we are not doing out-of-order hydration for the moment, we could bypass the `toVdom` processing of the client component inner nodes.

block-hydration-experiments.php
```diff
@@ -93,50 +102,70 @@ function bhe_block_wrapper($block_content, $block, $instance)
...
+ * TODO: use `WP_HTML_Tag_Processor` (see https://github.com/WordPress/gutenberg/pull/42485) once
+ * the API is released.
+ */
+function bhe_append_attributes($block_name, $block_content, $attributes)
+{
 	// Be aware that this pattern could not cover some edge cases.
-	$class_pattern     = '/class="\s*(?:[\w\s-]\s+)*' . $block_classname . '(?:\s+[\w-]+)*\s*"/';
-	$class_replacement = '$0 ' . $block_wrapper_attributes;
-	$block_content     = preg_replace( $class_pattern, $class_replacement, $block_content, 1 );
+	$pattern = '/^\s*<[^>]+/';
```
luisherranz, line 166: Ok, so this is now looking for the first tag opener, whatever it is, and it could be replaced with this, right?

```php
$w = new WP_HTML_Tag_Processor($block_content);
$w->next_tag();
foreach ($attributes as $key => $value) {
  $w->set_attribute($key, $value);
}
```

_With a bit more logic for the `style` and `class` attributes._

block-hydration-experiments.php
```diff
@@ -93,50 +102,70 @@ function bhe_block_wrapper($block_content, $block, $instance)
...
+
+add_filter('render_block_data', 'bhe_inner_blocks', 10, 3);
+
+/**
+ * Append attributes to the block wrapper element, which is assumed to be the first one.
+ *
+ * TODO: use `WP_HTML_Tag_Processor` (see https://github.com/WordPress/gutenberg/pull/42485) once
+ * the API is released.
+ */
+function bhe_append_attributes($block_name, $block_content, $attributes)
```
luisherranz, line 163: `$block_name` doesn't seem to be used anymore.

luisherranz (approved):
LGTM!! 

I'm going to merge this and resolve conflicts in https://github.com/WordPress/block-hydration-experiments/pull/74.

PR #78, opened 2022-09-29
⚛️ Use `Suspense` and `lazy` to Hydrate views that were not registered before calling `hydrate()`

DAreRodz:
_Coauthored with @c4rl0sbr4v0_

This PR replaces the `Map` we used to store the View components with an object with a similar API. The difference is that the method `get` returns a promise that resolves once the View component is registered.

With that Promise, we can use `Suspense` and `lazy` to pause hydration until the components are ready.

### DIFF
--- a/src/directives/wp-block-type.js
+++ b/src/directives/wp-block-type.js
@@ -66,8 +66,7 @@ directive('blockType', (wp, props, { vnode }) => {
 
 	if (!blockHydration) return;
 
-	// Use `lazy()` to pause hydration until the component is
-	// downloaded.
+	// Use `lazy()` to pause hydration until the component is registered.
 	const Component = lazy(() => blockViews.get(blockType));
 
 	if (blockUsesBlockContext.length) {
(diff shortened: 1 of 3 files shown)


luisherranz (approved):
LGTM 🙂👏

Note for the future: if we want to use [async](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-async), this code needs to be included in each bundle. To save some bytes on each bundle, I think we could find a way to move the promise-related logic to the hydration file instead. But this is perfect for now, thanks folks! 🎉

DAreRodz:
That was fast!! 😆 🚀

michalczaplinski:
Nice work 👏 

It's very similar to the work Bernie and me did in Woo-BHE to solve the problem of hydrating blocks when their views get registered too late:

https://github.com/woocommerce/woocommerce-blocks-hydration-experiments/commit/8edc7b2f2115c21fb11dcee86e92d24017b4c305

We ended up not using that code in the end because the problem because [there was a simpler solution](https://github.com/woocommerce/woocommerce-blocks-hydration-experiments/pull/21#issuecomment-1205248320)at the time.

DAreRodz:
Hey @michalczaplinski, I wasn't aware of that code! I see it's very similar, although it's using a `Proxy` to detect the Component's registration. Thanks for sharing it. 😊

Issue #80, opened 2022-10-14
Tracking issue: WordPress Directives Plugin 🎨


luisherranz:
This issue aims to list and keep track of the tasks related to the WordPress Directives Plugin experiment. The goal is to create an installable plugin that adds a set of basic directives and client-side navigations to any site.

_For this experiment's stress-testing, please check the [Tracking Issue: Performance and stress-testing vDOM Hydration 🧪](https://github.com/WordPress/block-hydration-experiments/issues/92)_

_For the Custom Elements or Directives Hydration experiment, please check their corresponding [Tracking Issues](https://github.com/WordPress/block-hydration-experiments/issues)._

The main branch of this experiment is [`main-wp-directives-plugin`](https://github.com/WordPress/block-hydration-experiments/tree/main-wp-directives-plugin). Make sure you select it when opening a pull request if it is related to this experiment.

Please proactively comment on this issue whenever you start working on a new task, get blocked, or finish a task, sharing as much detail and information as possible. Thanks!!

# High-level overview

- **Directives**

  - Finish the directives of the initial list - **In progress** 🚧

    Work on the directives is still reactive as we only work on the ones that we need for our experiments, demos, etc. The work on the Movies demo and Woo blocks should bring some focus to this task.

  - Add SSR for the ones that need it - **In progress** 🚧\

    _This is led by @ockham and @dmsnell._

    There's already initial support for the SSR of basic directive attributes and good progress on `wp-context`. The next challenge to solve is to figure out the APIs to access the inner HTML with the HTML (Tag) Processor.

  - Decide final names - **In progress** 🚧

    We are keeping a list of the decisions that need to be made here in the Tracking Issue.

- **Components**

  There hasn't been any progress on this, mostly due to the lack of need for components (directive tags). [@luisherranz opened an issue proposing their removal](https://github.com/WordPress/block-hydration-experiments/issues/152), as we couldn't find a single case where they add meaningful value and we might want to use them for other purposes in the future.

- **State Management**

  - Finish the Deep Signals implementation - [Finished](https://github.com/luisherranz/deepsignal) ✅
  - Decide on the recommended shape - **In progress** 🚧\
    _Same answer as in the "Directives - Decide final names"._
  - TypeScript support - **Not started** ⏸
  - Figure out if we want to expose some parts of the router and/or other configuration in the state ([example](https://github.com/c4rl0sbr4v0/wp-movies-demo/pull/15)) - **Not started** ⏸

[...]

PR #81, opened 2022-10-17
🎨 Add wp-link attribute to all links

SantosGuillamot:
I'm opening this Pull Request to "Add the wp-link to all the relative links on the page", one of the tasks of the [Tracking issue: WordPress Directives Plugin](https://github.com/WordPress/block-hydration-experiments/issues/80). This PR aims to just add that attribute in PHP, while all the logic will be handled within the directives in JS.

The first idea is to handle that in the [`render_block` filter](https://developer.wordpress.org/reference/hooks/render_block/) using the [WP_HTML_Tag_Processor](https://github.com/WordPress/gutenberg/pull/42485).

### DIFF
--- a/wp-directives.php
+++ b/wp-directives.php
@@ -40,9 +40,19 @@ add_action('init', 'wp_directives_register_scripts');
 
 function add_wp_link_attribute($block_content)
 {
+	$site_url = parse_url(get_site_url());
 	$w = new WP_HTML_Tag_Processor($block_content);
-	$w->next_tag('a');
-	$w->set_attribute('wp-link', 'true');
+	while ($w->next_tag('a')) {
+		$link = parse_url($w->get_attribute('href'));
+
+		if ($w->get_attribute('target') === '_blank') {
+			break;
+		}
+
+		if (is_null($link['host']) || ($link['host'] ===  $site_url['host'])) {
+			$w->set_attribute('wp-link', 'true');
+		}
+	};
 	return $w;
 }
 
--- a/wp-directives.php
+++ b/wp-directives.php
@@ -43,17 +43,16 @@ function add_wp_link_attribute($block_content)
 	$site_url = parse_url(get_site_url());
 	$w = new WP_HTML_Tag_Processor($block_content);
 	while ($w->next_tag('a')) {
-		$link = parse_url($w->get_attribute('href'));
-
 		if ($w->get_attribute('target') === '_blank') {
 			break;
 		}
 
+		$link = parse_url($w->get_attribute('href'));
 		if (is_null($link['host']) || ($link['host'] ===  $site_url['host'])) {
 			$w->set_attribute('wp-link', 'true');
 		}
 	};
-	return $w;
+	return (string) $w;
 }
 
 add_filter('render_block', 'add_wp_link_attribute', 10, 2);
--- a/wp-directives.php
+++ b/wp-directives.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Plugin Name:       wp-directives
  * Version:           0.1.0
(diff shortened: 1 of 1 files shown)


SantosGuillamot:
So far, I just added a single filter, but I encountered two problems:
* If there are two links in the same block, the `wp-link` attribute is only added to the first element. For example, a paragraph with two links:
```html
<p>First Link: <a wp-link="true" href="/">First Link</a>. Second Link: <a href="/sample-page">Second Link</a>.</p>
```
* It seems to be working fine with the current navigation block. But I quickly tested it with the [Experimental Navigation block using directives](https://github.com/WordPress/gutenberg/pull/44509) and it doesn't seem to apply.

I haven't triaged it yet, but any ideas why this is happening or how to solve it are welcome 🙂

luisherranz:
Oh, sorry. The HTML Tag Processor needs to be called recursively:

```php
$w = new WP_HTML_Walker( $html );
while ( $w->next_tag("a") ) {
  // ...
}
```

By the way, I've just noticed that in [one of the examples of the Make Core post](https://make.wordpress.org/core/2022/08/19/a-new-system-for-simply-and-reliably-updating-html-attributes/), they are using `$w->get_attribute("sr")`. Maybe it's already implemented. Could you please check it out?

SantosGuillamot:
> The HTML Tag Processor needs to be called recursively

That makes sense 😄 I've just changed it.

> By the way, I've just noticed that in [one of the examples of the Make Core post](https://make.wordpress.org/core/2022/08/19/a-new-system-for-simply-and-reliably-updating-html-attributes/), they are using $w->get_attribute("sr"). Maybe it's already implemented. Could you please check it out?

Yes, it seems implemented! 🙂 I added a check for the hostname and another one to exit the loop if `target === 'blank'`. Any more use cases where we shouldn't add the `wp-link` attribute?

I will try to take a look now at why this isn't working in the [navigation block using directives](https://github.com/WordPress/gutenberg/pull/44509).

luisherranz:
> Any more use cases where we shouldn't add the wp-link attribute?

We can add more later if we need to 🙂

wp-directives.php
```diff
@@ -36,3 +37,23 @@ function wp_directives_register_scripts()
...
+function add_wp_link_attribute($block_content)
+{
+	$site_url = parse_url(get_site_url());
+	$w = new WP_HTML_Tag_Processor($block_content);
+	while ($w->next_tag('a')) {
+		$link = parse_url($w->get_attribute('href'));
+
+		if ($w->get_attribute('target') === '_blank') {
+			break;
+		}
```
luisherranz, line 50: You can move this to the top of the `while` to avoid the cost of parsing the `href` if the target is `_blank`.

luisherranz, line 48: You can move this to the top of the `while` to avoid the cost of parsing the `href` if the target is `_blank`.

wp-directives.php
```diff
@@ -36,3 +37,23 @@ function wp_directives_register_scripts()
...
+
+		if (is_null($link['host']) || ($link['host'] ===  $site_url['host'])) {
+			$w->set_attribute('wp-link', 'true');
+		}
+	};
+	return $w;
```
luisherranz: I'm not 100% sure how is the final API, but I think it's safer to do this. If not, you'd be returning the `WP_HTML_Tag_Processor` instance instead:

```suggestion
	return (string) $w;
```

_The `(string) $variable` thing is similar to JavaScript's `.toString()`._

luisherranz (approved):
Looks good to me, Mario!

wp-directives.php
SantosGuillamot: Changed 🙂

wp-directives.php
SantosGuillamot: Changed 🙂

SantosGuillamot:
Thanks for the review! Changes addressed 🙂

PR #82, opened 2022-10-17
🎨 Add basic functionality

luisherranz:
This is basically a copy/paste from the [Navigation block experiment](https://github.com/WordPress/gutenberg/pull/44509).

I'll add the `wp-link` and router from the [Query Loop block experiment](https://github.com/WordPress/gutenberg/pull/44034) in a separate PR.

### DIFF
--- a/src/runtime/index.js
+++ b/src/runtime/index.js
@@ -1,1 +1,24 @@
-console.log('Runtime loaded correctly.');
+import { hydrate } from 'preact';
+import registerDirectives from './directives';
+import registerComponents from './components';
+import toVdom from './vdom';
+import { createRootFragment } from './utils';
+
+/**
+ * Initialize the initial vDOM.
+ */
+document.addEventListener('DOMContentLoaded', async () => {
+	registerDirectives();
+	registerComponents();
+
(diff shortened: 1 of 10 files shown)


src/runtime/index.js
```diff
@@ -1 +1,24 @@
...
+import { hydrate } from 'preact';
+import registerDirectives from './directives';
+import registerComponents from './components';
+import toVdom from './vdom';
+import { createRootFragment } from './utils';
+
+/**
+ * Initialize the initial vDOM.
+ */
+document.addEventListener('DOMContentLoaded', async () => {
```
DAreRodz, line 10: ~It seems that this function doesn't need to be `async`, does it?~

EDIT: I've just seen in the [next PR](https://github.com/WordPress/block-hydration-experiments/pull/83/files#diff-033681730ed7a6b0dc8ef504c5a7dfbfad95ce87165691be1f07533295989ab3R17) that you are using `await init()`, so there's nothing to change here.

DAreRodz (approved):
Looks good to me! 😄

luisherranz:
Thanks, David! 🎉

PR #83, opened 2022-10-17
🎨 Add basic router and wp link directive

luisherranz:
_This PR is branched on top of https://github.com/WordPress/block-hydration-experiments/pull/82._

This PR adds the remaining functionality: router + `wp-link` directive.

It's pretty much a copy/paste of the [Query Loop block](https://github.com/WordPress/gutenberg/pull/44034) and [Product Query block](https://github.com/woocommerce/woocommerce-blocks/pull/7187) experiments.

### DIFF
--- a/src/runtime/directives.js
+++ b/src/runtime/directives.js
@@ -1,7 +1,8 @@
-import { useContext, useMemo } from 'preact/hooks';
+import { useContext, useMemo, useEffect } from 'preact/hooks';
 import { useSignalEffect } from '@preact/signals';
 import { directive } from './hooks';
 import { deepSignal } from './deepsignal';
+import { prefetch, navigate } from './router';
 import { getCallback } from './utils';
 
 const raf = window.requestAnimationFrame;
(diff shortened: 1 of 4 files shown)


SantosGuillamot:
Not sure if I'm doing something wrong but it seems this PR breaks the `WP Admin`. I have trouble navigating between pages and at some point, I am logged out automatically. I guess we will need to add something to prevent this from happening in the `WP Admin`?

EDIT: Oh, it is [this code](https://github.com/WordPress/block-hydration-experiments/blob/basic-router-and-wp-link-directive/src/runtime/index.js#L12-L15) to test it out and it should be fixed once we add the wp-link attribute in PHP in https://github.com/WordPress/block-hydration-experiments/pull/81?

DAreRodz (approved):
I haven't tested it, but the code looks great. 💯

SantosGuillamot:
> EDIT: Oh, it is [this code](https://github.com/WordPress/block-hydration-experiments/blob/basic-router-and-wp-link-directive/src/runtime/index.js#L12-L15) to test it out and it should be fixed once we add the wp-link attribute in PHP in https://github.com/WordPress/block-hydration-experiments/pull/81?

Yes, that was the issue I was facing. I have just merged https://github.com/WordPress/block-hydration-experiments/pull/81 so I assume we can remove [this code](https://github.com/WordPress/block-hydration-experiments/blob/basic-router-and-wp-link-directive/src/runtime/index.js#L12-L15) right? However, in that PR I am adding `wp-link="true"` and we are expecting something like `wp-link="{ "prefetch": true }"`. Should we transform `wp-link="true"` into `wp-link="{ "prefetch": true }"` in the directive or in PHP?

Apart from that, I made the changes manually and it works great! 🎉 👏

luisherranz:
I've removed the manual `wp-link` injection and enqueued the scripts only on the frontend (my fault).

> Should we transform wp-link="true" into wp-link="{ "prefetch": true }" in the directive or in PHP?

Yeah, but first we need to consider what prefetching strategy we want to enable here.

PR #84, opened 2022-10-17
🎨 Remove comments and preserve scripts

### LINKED ISSUE #79: Preact: Hydration fails when there are HTML comments
We found what I think is a bug in Preact's hydration. By default, if Preact finds HTML comments during the hydration, it removes them. The problem is that while doing so, it also recreates (removes and adds again) the next DOM node.

I've reproduced the issue in this repository: https://github.com/luisherranz/preact-comments-hydration-bug

I've also checked it in Preact's own tests, and it's also happening there: https://github.com/luisherranz/preact/commit/004d25b0673556e56c6bcdf3ce87034865a55bfb

I've recorded a video to show the repository and tests:

<a href="https://www.loom.com/share/89dbe2ac14164bb9a35457cd27e10cce">
    
  </a>

https://www.loom.com/share/89dbe2ac14164bb9a35457cd27e10cce

luisherranz:
I don't remember why we removed the script's children in the first place. We should investigate that, so let's preserve them again and see when they fail.

I've also added logic to remove the comments because [Preact doesn't support them during the hydration](https://github.com/WordPress/block-hydration-experiments/issues/79). We may need to add them back after the hydration if sites rely on them, but for now, I'd keep them removed.

### DIFF
--- a/src/runtime/vdom.js
+++ b/src/runtime/vdom.js
@@ -1,18 +1,13 @@
-/**
- * External dependencies
- */
 import { h } from 'preact';
 
 // Recursive function that transfoms a DOM tree into vDOM.
 export default function toVdom(node) {
 	const props = {};
-	const attributes = node.attributes;
+	const { attributes, childNodes } = node;
 	const wpDirectives = {};
 	let hasWpDirectives = false;
 
 	if (node.nodeType === 3) return node.data;
-	if (node.nodeType === 8) return null;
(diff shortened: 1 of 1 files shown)


SantosGuillamot (approved):
LGTM! 🙂 Although I'm not 100% sure how to test it.

luisherranz:
We'll keep testing it, don't worry 🙂

PR #85, opened 2022-10-19
🎨 Support CSS in client-side navigations

luisherranz:
I've added support for loading the CSS of the head during the client-side navigations.

I finally opted for not using vDOM diffing, and I also opted for using a `<style>` for `<link rel="stylesheet">` to make sure we do the navigation (body and CSS) in a single step and avoid any FOUC. I think it's simpler this way, but happy to hear other opinions.

<a href="https://www.loom.com/share/827311414ffd4acdacb2b675b09050fa">
    
  </a>

https://www.loom.com/share/827311414ffd4acdacb2b675b09050fa

@darerodz: please check that this solves your problems with CSS during the navigations.

### DIFF
--- a/src/runtime/router.js
+++ b/src/runtime/router.js
@@ -6,7 +6,8 @@ import { createRootFragment } from './utils';
 let rootFragment;
 
 // The cache of visited and prefetched pages.
-export const pages = new Map();
+const pages = new Map();
+const stylesheets = new Map();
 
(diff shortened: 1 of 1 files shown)


DAreRodz (approved):
It looks great. Simple and effective. 💯

luisherranz:
Thanks, David.

PR #86, opened 2022-10-19
🎨 Add meta tag to opt-in to client-side navigations

luisherranz:
I've added a mechanism to opt-in to client-side navigations using a metatag, so navigation can do a fallback after receiving the HTML of a page that is not client-side navigation friendly.

I left them activated by default for now, although I want to add an admin screen to toggle them on and off.

<a href="https://www.loom.com/share/32bfdeb904e04c7e933474744e11be81">
    
  </a>

https://www.loom.com/share/32bfdeb904e04c7e933474744e11be81

### DIFF
--- a/wp-directives.php
+++ b/wp-directives.php
@@ -34,10 +34,9 @@ function wp_directives_register_scripts()
 	// conditionally enqueue directives later.
 	wp_enqueue_script('wp-directive-runtime');
 }
-
 add_action('wp_enqueue_scripts', 'wp_directives_register_scripts');
 
-function add_wp_link_attribute($block_content)
+function wp_directives_add_wp_link_attribute($block_content)
 {
 	$site_url = parse_url(get_site_url());
 	$w = new WP_HTML_Tag_Processor($block_content);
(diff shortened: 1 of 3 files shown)


wp-directives.php
```diff
@@ -53,5 +52,15 @@ function add_wp_link_attribute($block_content)
...
+{
+	if (apply_filters('client_side_transitions', false)) {
+		echo '<meta itemprop="wp-client-side-transitions" content="active">';
+	}
+}
+add_action('wp_head', 'wp_directives_client_site_transitions_meta_tag', 10, 0);
 
-add_filter('render_block', 'add_wp_link_attribute', 10, 2);
+/* User code */
+add_filter('client_side_transitions', '__return_true');
```
SantosGuillamot, line 66: If we are planning to add a setting [here](https://github.com/WordPress/block-hydration-experiments/pull/88) to enable Client Side Transitions, or if the default will be false, should we remove this line?

EDIT: Nevermind, I just saw you are removing it in the other PR.

SantosGuillamot (approved):
Looks great! Thanks Luis 🙂

luisherranz:
Thanks, Mario!

PR #87, opened 2022-10-19
🎨 Add `prefetch` to all links and prevent scroll in pagination

SantosGuillamot:
I've added the `prefetch` property to all the links and `scroll=false` to all the pagination links. While doing so we realized that the HTML of the parent blocks contain the children, so we were parsing the HTML multiple times. For that, we change the filters to only trigger for the Query Loop and the Template Parts until we figure out a better solution.

### DIFF
--- a/wp-directives.php
+++ b/wp-directives.php
@@ -47,8 +47,8 @@ function add_wp_link_attribute($block_content)
 		}
 
 		$link = parse_url($w->get_attribute('href'));
-		$classes = $w->get_attribute('class');
 		if (!isset($link['host']) || $link['host'] === $site_url['host']) {
+			$classes = $w->get_attribute('class');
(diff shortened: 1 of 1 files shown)


wp-directives.php
```diff
@@ -47,11 +47,18 @@ function add_wp_link_attribute($block_content)
 		}
 
 		$link = parse_url($w->get_attribute('href'));
+		$classes = $w->get_attribute('class');
```
luisherranz: It's a minimal optimization, but you can move this getter inside the `if`.

luisherranz (approved):
Lovely.

wp-directives.php
SantosGuillamot: Totally right! I've just changed it 🙂

PR #88, opened 2022-10-19
🎨 Admin for client side navigations

luisherranz:
It's just an admin page to opt-in to the client-side navigations.

I did it on top of https://github.com/WordPress/block-hydration-experiments/pull/86, so we need to merge that one first.

Screenshot 2022-10-19 at 20 07 20

**This is meant for testing only. It doesn't mean we intend to add an option to WordPress**.

If you had the plugin installed before, please deactivate and activate the plugin again so it creates the default settings.

### DIFF
--- a/src/admin/admin-page.php
+++ b/src/admin/admin-page.php
@@ -1,0 +1,82 @@
+<?php
+
+function wp_directives_register_menu()
+{
+	add_options_page(
+		'WP Directives',
+		'WP Directives',
+		'manage_options',
+		'wp-directives-plugin',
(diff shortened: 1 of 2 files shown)


SantosGuillamot (approved):
LGTM! 🙂 Btw, if in the future we decide to start only with the filter and not the Admin page, I've realized that we can ask to use plugins like [Code Snippets](https://wordpress.org/plugins/code-snippets/) for those that can't modify the code.

luisherranz:
Indeed 🙂 

Thanks, Mario!

PR #91, opened 2022-10-24
🎨 WP Directives plugin - Stress testing in real sites

michalczaplinski:
> **Note**: If someone wants to help with the tests, please send me a DM on Slack (wordpress.slack.com) and I'll send you a link with the `csv` files that contain the list of sites.

---

This PR adds a script that runs some stress tests against multiple live WordPress pages and uses the MutationObserver to check for DOM modifications.

https://user-images.githubusercontent.com/5417266/198370176-109b0977-cbde-47c4-8edc-56144132495b.mp4


michalczaplinski:
The latest results are in the `results_first_100.json`file.

I'm currently focused on this:
- [x] Parallelize the tests so that we can test more sites quickly (currently they take a few mins for just the first 100 sites on my M1 Macbook.
- [x] Save the data into an sqlite database instead of into a JSON file. Looking at JSON files gets unwieldy at scale. This way we can:
   -  Run the tests for the first 100 pages, save them to database and push to github. Then, we can resume testing next day without re-running the tests for the same first 100 sites.
   - We can use a tool like https://datasette.io/ to quickly look at the results.
   - We can keep track of which sites were tested and which ones resulted in an error. Not impossible with JSON but a bit harder to keep track of.

SantosGuillamot:
This looks great Michal! I made some Pull Requests with potential small improvements. Maybe some of them don't make much sense, just let me know 🙂 

- https://github.com/WordPress/block-hydration-experiments/pull/95
- https://github.com/WordPress/block-hydration-experiments/pull/96
- https://github.com/WordPress/block-hydration-experiments/pull/97/

package.json
```diff
@@ -6,7 +6,9 @@
 	"main": "build/index.js",
 	"scripts": {
 		"build": "webpack --mode=production",
+		"build:benchmark": "esbuild benchmark/hydrationScript.js --bundle --outfile=build/hydrationScriptForTesting.js",
 		"start": "webpack --mode=development --watch",
+		"benchmark": "node ./benchmark/index.mjs",
```
cbravobernal: Would be better to run first the build so it doesn't fail if we didn't run it before?
Co-reviewed with @DAreRodz 

```"benchmark": "npm run build:benchmark && node ./benchmark/index.mjs"```

**Comment by michalczaplinski** [comment]
Those are some of the immediate next steps:

- [ ] Update the stress testing to verify that the hydration has completed correctly:
  - [x] Wrap the hydration code in some trycatch. If there is an error thrown during hydration, save it to the database.
  - [ ] We can also check if `window.document.documentElement.__k.__k` object is present which it should after hydration completes successfully
- [ ] Investigate why the our hydration fails on some React (Next.js) sites. The sites that we know were failing were [chegg.com](http://chegg.com/) and [bloomberg.com/professional/](https://www.bloomberg.com/professional/).
- [ ] Manually test some sites by copy-pasting the [`manual-testing-script.js`](https://github.com/WordPress/block-hydration-experiments/blob/1f69c518ca6cc4aec7e034af242ba656ebd390d2/benchmark/manual-testing-script.js#L1-L3) in the browser's console or by using the [proxy](https://github.com/WordPress/block-hydration-experiments/pull/98).
- [ ] Run the tests on more sites. We estimate that [we should run the tests on hundreds of thousands of sites]

File: package.json (hunk 1 above)
Comment (michalczaplinski) [reply package.json]: Indeed!

**Comment by luisherranz** [comment]
I've started testing `1.csv`.

EDIT: I finished with four positives:
- 31shobo.com
- libertaegiustizia.it
- rivistasegno.eu
- cleofas.com.br

**Comment by SantosGuillamot** [comment]
I've started testing `2.csv`.

EDIT: Finished with one positive as well:

* nigloland.fr

**Comment by cbravobernal** [comment]
I've started testing `3.csv`

Results so far:
17632 pages analyzed
17521 success
108 timeoutError
233 error
40 NULL

3 Positives:

- http://gekkanbijutsu.co.jp/: Comments are being removed. The website is working as expected.
- valoda.lv: false positive, there is a slider mutating the DOM all the time. Our script is not interfering with the auto-slide feature. The website is working as expected.
- informatorenavale.it: Comments are being removed. The website is working as expected.

**Comment by luisherranz** [comment]
@michalczaplinski, with the current tests, is it possible that the hydration may happen before [the `DOMContentLoaded` event](https://github.com/WordPress/block-hydration-experiments/blob/main-wp-directives-plugin/src/runtime/index.js#L8)? Or said with different words: is it possible that [page.evaluate](https://playwright.dev/docs/evaluating) may run before the DOM is ready? 

I saw a few positives that I cannot reproduce manually, and I'm trying to figure out why.

**Comment by luisherranz** [comment]
I'm going to try with:

```js
await page.goto(`http://${url}`, {
  waitUntil: "domcontentloaded", // <-- instead of network-idle
  timeout: 60000,
});
```

luisherranz:
I've started testing `4.csv`.

EDIT: Finished with one positive:

- nigloland.fr

I haven't looked into it, but I can reproduce the nigloland.fr mutations using `manual-testing-script.js` in the console.

SantosGuillamot:
>EDIT: Finished with one positive:
>
> * nigloland.fr

Is it possible that you have tested `2.csv` instead of `4.csv`? `nigloland.fr` was [the only positive in the second file](https://github.com/WordPress/block-hydration-experiments/pull/91#issuecomment-1303193436), and it doesn't seem to be in [the fourth one](https://github.com/WordPress/block-hydration-experiments/blob/perf-and-stress-testing/benchmark/data/4.csv).

SantosGuillamot:
I've started testing `10.csv`.

EDIT: Finished with three positives:

* congresos-medicos.com: Doesn't seem to be a WordPress site.
* theglassescompany.co.uk: Doesn't seem to be a WordPress site.
* nedsahin.com: I wasn't able to reproduce it.

SantosGuillamot:
I've started testing `5.csv`.

EDIT: Finished with one positive:

* giocodellanno.it: I wasn't able to reproduce it.

luisherranz:
> Is it possible that you have tested 2.csv instead of 4.csv? nigloland.fr was the only positive in the second file, and it doesn't seem to be in the fourth one.

Oh, yeah, I totally did that.

Starting again with `4.csv`😅

DAreRodz:
Just in case anyone is planning to run some tests, I'll do it with `9.csv` and, if I have some extra time, also with `8.csv`. ☝️

SantosGuillamot:
It seems that in all the sites that were failing but we weren't able to reproduce manually, there is a library ([Mootools](https://mootools.net/) or [prototype](http://prototypejs.org/) for example) that is overwriting toJSON function of the Array prototype. This means that we aren't iterating properly the list of added/removed nodes, saving wrong results.

We can assume these are false positives and it would be something to fix in our testing (if we want), it shouldn't affect the hydration script.

This is the list of sites affected by this so far:
* 31shobo.com
* libertaegiustizia.it
* rivistasegno.eu
* cleofas.com.br
* congresos-medicos.com
* theglassescompany.co.uk
* nedsahin.com
* giocodellanno.it

With this, only **nigloland.fr** is having real mutations so far (investigating it right now).

Props to @DAreRodz for discovering it.

SantosGuillamot:
I was testing **nigloland.fr**, the only positive we have right now. This is what I found so far:

**Nodes with issues**

Checking out the modifications, it seems we are just removing and adding the same HTML. The nodes affected are the ones returned by `document.querySelectorAll(".wpsbc-legend-item p")`. They are some paragraphs at the bottom of the page, just below the calendar:

Screenshot 2022-11-14 at 14 04 12

**Potential script causing the issue**

I've been taking a look at the scripts that could affect these nodes and it seems that the one causing the modifications is this one:

```html
<script type='text/javascript' src='https://www.nigloland.fr/wp-content/themes/nigloland/assets/js/app.js' id='app-js'></script>
```

I've tested the site removing this script using @DAreRodz [proxy](https://github.com/WordPress/block-hydration-experiments/pull/98), and the modifications aren't happening when I paste the [manual-testing-script](https://github.com/WordPress/block-hydration-experiments/blob/perf-and-stress-testing/benchmark/manual-testing-script.js).

Taking a deeper look at [the mentioned script](https://www.nigloland.fr/wp-content/themes/nigloland/assets/js/app.js), the part interacting with those nodes is the following one:

```js
function changeLegend(){
	//console.log("Changelegend");
	$(".page-template-page-homepage .wpsbc-legend-item").each(function() {
		var txt = $(this).find('p').html();
		var txtReplace = txt.split('<br>');
		$(this).find('p').html(txtReplace[0]);
	});
}

changeLegend();

$('.wpsbc-calendar-1').bind("DOMSubtreeModified",function(){
  	$('.wpsbc-calendar-1').unbind("DOMSubtreeModified");
  	setTimeout(changeLegend(), 5000);
});
```
I'm not 100% sure this is the code causing the issue. Any idea if this could be the cause of the hydration failing? Could it be the `DOMSubtreeModified`? It seems it was [deprecated in favor of the Mutation Observer API](https://docs.w3cub.com/dom_events/domsubtreemodified), so maybe it is colliding somehow?

luisherranz:
Have you checked if these mutations also occur without our hydration (injecting only the `MutationObserver`)?

By the way, at this point, I would define two types of positives:

- Positives: there are mutations in the DOM, and the resulting HTML is not the same.
- Soft positives: there are mutations in the DOM, but the resulting HTML is the same.

We need to keep investigating both, but soft positives are not as dangerous.

SantosGuillamot:
> Have you checked if these mutations also occur without our hydration (injecting only the MutationObserver)?

I've just tried commenting [the `window.__runHydration();` line](https://github.com/WordPress/block-hydration-experiments/blob/perf-and-stress-testing/benchmark/manual-testing-script.js#L759) and it doesn't log any mutations, so I assume it doesn't occur without our hydration.

>By the way, at this point, I would define two types of positives:
>
>* Positives: there are mutations in the DOM, and the resulting HTML is not the same.
>* Soft positives: there are mutations in the DOM, but the resulting HTML is the same.

Makes sense 🙂

SantosGuillamot:
>Taking a deeper look at [the mentioned script](https://www.nigloland.fr/wp-content/themes/nigloland/assets/js/app.js), the >part interacting with those nodes is the following one:

I've been running a few more checks and it really seems it is this code causing the issues. If I run the script without that part the mutations aren't registered. It seems it only causes mutations when both the line `$(this).find('p').html(txtReplace[0]);` and the `DOMSubtreeModified` part are active. This is what I observed:

**Only case where it causes issues**
When both parts are active
```js
function changeLegend(){
	$(".page-template-page-homepage .wpsbc-legend-item").each(function() {
		var txt = $(this).find('p').html();
		var txtReplace = txt.split('<br>');
		$(this).find('p').html(txtReplace[0]);
	});
}

changeLegend();

$('.wpsbc-calendar-1').bind("DOMSubtreeModified",function(){
  	$('.wpsbc-calendar-1').unbind("DOMSubtreeModified");
  	setTimeout(changeLegend(), 5000);
});
```

**Cases where it works fine**
1. `console.log` instead of `$(this).find()`

```js
function changeLegend(){
	$(".page-template-page-homepage .wpsbc-legend-item").each(function() {
		console.log("Test");
	});
}

changeLegend();

$('.wpsbc-calendar-1').bind("DOMSubtreeModified",function(){
  	$('.wpsbc-calendar-1').unbind("DOMSubtreeModified");
  	setTimeout(changeLegend(), 5000);
});
```

2. Commenting the `DOMSubtreeModified`

```js
function changeLegend(){
	$(".page-template-page-homepage .wpsbc-legend-item").each(function() {
		console.log("Test");
		var txt = $(this).find('p').html();
		var txtReplace = txt.split('<br>');
		$(this).find('p').html(txtReplace[0]);
	});
}

changeLegend();

// $('.wpsbc-calendar-1').bind("DOMSubtreeModified",function(){
//   	$('.wpsbc-calendar-1').unbind("DOMSubtreeModified");
//   	setTimeout(changeLegend(), 5000);
// });
```

I have no idea why this is happening, but according to this jQuery issue, [`$().find()` triggers DOMSubTreeModified events](https://bugs.jquery.com/ticket/8046). It seems `$().find()` changes the id of nodes as it traverses. Could this be the problem?

Anyway, this is a soft positive which seems to be a pretty edge case with an API that is already deprecated. Maybe we should just ignore this one?

luisherranz:
> Anyway, this is a soft positive which seems to be a pretty edge case with an API that is already deprecated. Maybe we should just ignore this one?

I agree. We can add this to the list of known incompatibilities and keep going on 🙂

luisherranz:
> I've just tried commenting the window.__runHydration(); line and it doesn't log any mutations, so I assume it doesn't occur without our hydration.

Could you try commenting `toVdom()` and `hydrate()` separately? I would like to know which one is affecting the `DOMSubtreeModified`.

luisherranz:
By the way, [this is the list of plugins that are supposed to be using `DOMSubtreeModified`](https://wpdirectory.net/search/01GHX8AABC0Z151YTRTHZSQK14). Maybe we could test the most important ones. 

_It claims that Jetpack is using it, but I am not able to find it in [their codebase](https://github.com/Automattic/jetpack/search?q=DOMSubtreeModified)._

SantosGuillamot:
>Could you try commenting toVdom() and hydrate() separately? I would like to know which one is affecting the DOMSubtreeModified.

Not 100% sure that I'm doing it right, but it seems it is `toVdom()` the one affecting. I'm commenting [these two lines](https://github.com/WordPress/block-hydration-experiments/blob/perf-and-stress-testing/benchmark/manual-testing-script.js#L706-L707).

1. Commenting `toVdom()` and using `document.body` directly doesn't return the mutations. Maybe it is because I am using `document.body` and I should be using something different?

```js
// const body = toVdom(document.body);
S(document.body, rootFragment);
```

2. Commenting the `hydrate()` returns the mutations:
```js
toVdom(document.body);
// S(body, rootFragment);
```

Btw, basically what I'm doing is pasting the manual-testing-script file in [nigloland.fr](https://www.nigloland.fr/) console in the browser. Not sure if I should be doing something different.

DAreRodz:
> It claims that Jetpack is using it, but I am not able to find it in [their codebase](https://github.com/Automattic/jetpack/search?q=DOMSubtreeModified).

I took a quick look, and I was able to find `DOMSubtreeModified` in [`resize-observer-polyfill`](https://github.com/que-etc/resize-observer-polyfill/blob/4a148452494a155656e4d04b9732b5914896b2e0/src/ResizeObserverController.js#L171), which is a dependency of the Jetpack's Tiled Gallery block (see [view.js](https://github.com/Automattic/jetpack/blob/b313bd253afdf919adf7a71a866d0ab85b3a7bda/projects/plugins/jetpack/extensions/blocks/tiled-gallery/view.js#L3)).

It seems to be used only when the browser does not support `MutationObserver`, though.

luisherranz:
> basically what I'm doing is pasting the manual-testing-script file in nigloland.fr console in the browser

Ohhh, right. Sorry, I didn't realize that. I'll do that testing myself 🙂

> I took a quick look, and I was able to find DOMSubtreeModified in resize-observer-polyfill,

Good catch!

DAreRodz:
@SantosGuillamot, what you observed seems right to me. The `toVdom()` function modifies the DOM as it removes the HTML comments while building the vDOM.

Maybe the script [nigloland.fr](https://www.nigloland.fr/) is using detects those mutations with `DOMSubtreeModified` and makes extra mutations that are detected by our `MutationObserver`.

SantosGuillamot:
>Ohhh, right. Sorry, I didn't realize that. I'll do that testing myself 🙂

Oh, I didn't mean that! I just wanting to make sure I was testing it correctly 🙂 

>Maybe the script [nigloland.fr](https://www.nigloland.fr/) is using detects those mutations with DOMSubtreeModified and makes extra mutations that are detected by our MutationObserver.

The only related thing I found was this jQuery issue, [$().find() triggers DOMSubTreeModified events](https://bugs.jquery.com/ticket/8046). It seems `$().find()` changes the id of nodes as it traverses. But I don't fully understand it to be honest.

luisherranz:
> Maybe the script nigloland.fr is using detects those mutations with DOMSubtreeModified and makes extra mutations that are detected by our MutationObserver.

You're a genius! That's precisely what is happening 😄

I just ran this to remove the comments before the `manual-testing-script.js` file, and it doesn't log any mutations:

```js
const removeComments = (node) => {
  if (node.nodeType === 8) node.remove();
  [].map.call(node.childNodes, removeComments);
};
removeComments(document.body);
```

SantosGuillamot:
I created a couple PRs to improve the testing:
* https://github.com/WordPress/block-hydration-experiments/pull/102
* https://github.com/WordPress/block-hydration-experiments/pull/101

cbravobernal:
[Updated my comment](https://github.com/WordPress/block-hydration-experiments/pull/91#issuecomment-1303212620) with the final results, TLDR: good to go!

luisherranz:
> 3 Positives:
> 
> - http://gekkanbijutsu.co.jp/: Comments are being removed. The website is working as expected.
> - valoda.lv: false positive, there is a slider mutating the DOM all the time. Our script is not interfering with the auto-slide feature. The website is working as expected.
> - informatorenavale.it: Comments are being removed. The website is working as expected.

@c4rl0sbr4v0 two questions from your results:

- Do you know why the script is still reporting removed comments? Shouldn't that be already taken into account?
- Do you know why the script is picking mutations from the auto-slide if the `toVom + hydrate` is synchronous?

DAreRodz:
I've been testing `8.csv` and `9.csv` in parallel with the latest changes (including https://github.com/WordPress/block-hydration-experiments/pull/101 and https://github.com/WordPress/block-hydration-experiments/pull/102 that haven't been merged yet). I think some of the numbers include previously tested sites - those already present in the committed DB - but it's a small amount that we can ignore.

Below are the results so far.

| result | count | percent (%) | cloudflare (skipped) |
| --- | --: | --: | --: |
| all | 8436 | 100 | 702 |
| success | 6772 | 80.28 | 3 |
| NULL | 735 | 8.71 | 699 |
| timeoutError | 735 | 8.71 | 0 |
| error | 147 | 1.74 | 0 |
| 403error | 47 | 0.56 | 0 |

No positives yet. 🙂

cbravobernal:
> Do you know why the script is still reporting removed comments? Shouldn't that be already taken into account?
Do you know why the script is picking mutations from the auto-slide if the toVom + hydrate is synchronous?

After updating the branch with the last changes, and running the script only on the positives. Comments mutations disappeared and the script is not picking the mutations of the auto-slide as expected. 🤷‍♂️ 0 positives in `3.csv`

luisherranz:
Ok. Thanks, Carlos!

DAreRodz:
I'm currently working on integrating the [proxy experiment](https://github.com/WordPress/block-hydration-experiments/pull/98) in this PR, and while doing some testing, I noticed that if the `window.__runHydration()` call fails, the result saved in DB is `success`. You can try the following diff:

```diff
diff --git a/benchmark/index.mjs b/benchmark/index.mjs
index 7297439..f2a956f 100644
--- a/benchmark/index.mjs
+++ b/benchmark/index.mjs
@@ -279,6 +279,7 @@ async function runScript(wordPressPage, url, browser) {
 
 			try {
 				window.__runHydration();
+				throw 'FLAMES!';
 			} catch (error) {
 				console.log('error', JSON.stringify(error));
 			}
```

I think the expected behavior is to recognize the error and save `hydrationError` instead of `success`.

https://github.com/WordPress/block-hydration-experiments/blob/d77c6bba108cb7f13c450c12ed87065e5f3d6062/benchmark/index.mjs#L180-L186

Any idea of why is this happening? 🤔

SantosGuillamot:
> Any idea of why is this happening? 🤔

Not too familiar with `try...catch` but, if the `window.__runHydration()` function is already inside a `try...catch`, shouldn't we just use the function without another `try...catch`?

And, if for any reason, we need the `runHydration()` inside a new `try...catch`, shouldn't we throw the error again in the catch? Something like:

```
try {
    window.__runHydration();
} catch (error) {
    console.log('error', JSON.stringify(error));
    throw error;
}
```

I guess we could even throw something like `"hydrationError"` and register that in the DB to know in which sites the hydration is failing.

As I've said I haven't a clear understanding of `try...catch`, but just using the `runHydration()` function without a new `try...catch`, or throwing again the error, seems to work.

SantosGuillamot:
I've been testing another bunch of sites (`6.csv`). I found 2 positives (I haven't triaged them yet properly):
* [tiens-donc.com](http://tiens-donc.com/)
* [yourcontentempire.com](http://yourcontentempire.com/)

It's a bit weird because, taking a quick look at it, it seems that [`attributes.length` inside the `toVdom` function](https://github.com/WordPress/block-hydration-experiments/blob/perf-and-stress-testing/benchmark/manual-testing-script.js#L646) is not defined and it never exits the `for` loop. This is a quick video showing it.

This happens in both sites. The first one takes a bit longer to show the same mutation multiple times. The second one is faster. This is a quick video showing it (you can use 2x speed):

https://www.loom.com/share/4c9317388aca4b7ea4726e6d7658de67

Apart from that, I was a bit worried about [this comment David made](https://github.com/WordPress/block-hydration-experiments/pull/91#issuecomment-1318990967), in case the hydration was failing in some sites, so I added a check to register it in the database if that happened. It seems it didn't fail in any of the sites tested here.

These are the results:

| result | count | percent (%) | cloudflare (skipped) |
| --- | --: | --: | --: |
| all | 16926 | 100 | 1902|
| success | 14408 | 85.12 | 0 |
| NULL | 1902 | 11.24 | 1902 |
| timeoutError | 266 | 1.57 | 0 |
| error | 260 | 1.54 | 0 |
| 403error | 90 | 0.53 | 0 |

SantosGuillamot:
More sites testing (`7.csv`). There is 1 positive (not triaged yet) with a similar case to [the previous comment](https://github.com/WordPress/block-hydration-experiments/pull/91#issuecomment-1320886841):
* 24life.ro

It returns an error in the `toVdom` function and it never exits the `for` loop. In this case, it return the following error referring to [this line](https://github.com/WordPress/block-hydration-experiments/blob/perf-and-stress-testing/benchmark/manual-testing-script.js#L654):

`TypeError: object null is not iterable (cannot read property`.

These are the results:

| result | count | percent (%) | cloudflare (skipped) |
| --- | --: | --: | --: |
| all | 17437 | 100 | 1828 |
| success | 15008 | 86.07 | 0 |
| NULL | 1828 | 10.48 | 1828 |
| timeoutError | 274 | 1.57 | 0 |
| error | 263 | 1.50 | 0 |
| 403error | 64 | 0.37 | 0 |

DAreRodz:
@SantosGuillamot, to be sure, those sites you mentioned where `toVdom` execution fails, do they have `hydrationError` as the result? I think they should, but I don't see any in the results you shared.

DAreRodz:
> It's a bit weird because, taking a quick look at it, it seems that [`attributes.length` inside the `toVdom` function](https://github.com/WordPress/block-hydration-experiments/blob/perf-and-stress-testing/benchmark/manual-testing-script.js#L646) is not defined and it never exits the `for` loop. This is a quick video showing it.

@SantosGuillamot, I watched the video, and I think the problem is this:
1. `toVdom` fails and throws an exception.
2. That exception is not caught, so [`window.__runHydration`](https://github.com/WordPress/block-hydration-experiments/blob/perf-and-stress-testing/benchmark/manual-testing-script.js#L759) and all the code after that also fails.
3. Therefore, the observer is not disconnected, and it keeps registering mutations that those websites normally do.

We need to find out why `toVdom` is failing. I'm afraid it could be failing as well in some sites that don't mutate anything, and that we haven't detected. It should be a type of node that doesn't have attributes, or something we're passing to `toVdom` that is not a node.

EDIT: the problem I mentioned above only applies to the manual testing script, which is the one you use in the video.

DAreRodz:
I created this PR to fix the registering of `hydrationError` cases.

- https://github.com/WordPress/block-hydration-experiments/pull/106

DAreRodz:
> More sites testing (`7.csv`). There is 1 positive (not triaged yet) with a similar case to [the previous comment](https://github.com/WordPress/block-hydration-experiments/pull/91#issuecomment-1320886841):
> 
> * 24life.ro
> 
> It returns an error in the `toVdom` function and it never exits the `for` loop. In this case, it return the following error referring to [this line](https://github.com/WordPress/block-hydration-experiments/blob/perf-and-stress-testing/benchmark/manual-testing-script.js#L654):
> 
> `TypeError: object null is not iterable (cannot read property`.

I think this one could be related to what @michalczaplinski mentioned in https://github.com/WordPress/block-hydration-experiments/pull/99

SantosGuillamot:
I took a quick look at the three positives we have left, and this is what I found so far:
* The infinite loop of registering mutations seems to be solved with @DAreRodz PR: https://github.com/WordPress/block-hydration-experiments/pull/106
* 24life.ro seems to be fixed in @michalczaplinski PR #99 as @DAreRodz suggested.
* [tiens-donc.com](http://tiens-donc.com/) & [yourcontentempire.com](http://yourcontentempire.com/) have issues with some `<style>` tags included in the body.

Both of them have a `<style>` element with `<! [CDATA[...]]>` inside it. They are in `<svg>` elements. And it seems we are removing that part. Something similar to this:

```html
<svg ...>
    <style>
        <![CDATA[
            .st0{fill:#FFFFFF;}
            .st1{fill:#070707;}
        ]]>
    </style>
</svg>
```

Any idea how to approach it?

DAreRodz:
> Both of them have a `<style>` element with `<! [CDATA[...]]>` inside it. They are in `<svg>` elements. And it seems we are removing that part. Something similar to this:
> 
> ```
> <svg ...>
>     <style>
>         <![CDATA[
>             .st0{fill:#FFFFFF;}
>             .st1{fill:#070707;}
>         ]]>
>     </style>
> </svg>
> ```
> 
> Any idea how to approach it?

AFAIK, broswers tread `CDATA` tags as comments (it's actually not valid HTML but XML). I would expect our `toVdom` function to remove them like any other comment but I don't think those mutations should appear. Maybe we're not filtering them from the `MutationObserver` output? 🤔

In any case, I would say this kind of mutations is expected and not important. 🙂

SantosGuillamot:
> Maybe we're not filtering them from the MutationObserver output? 🤔

The `nodeName` of the mutations registered is `#text`, so we aren't considering it a comment because [this line is not true](https://github.com/WordPress/block-hydration-experiments/blob/perf-and-stress-testing/benchmark/index.mjs#L233).

> In any case, I would say this kind of mutations is expected and not important. 🙂

Okay, that's great! Removing the `CDATA` doesn't affect the style of the `svg`?

luisherranz:
I've opened an issue to discuss what we should do with `CDATA`:

- https://github.com/WordPress/block-hydration-experiments/issues/109

luisherranz:
> 24life.ro seems to be fixed in @michalczaplinski PR Check if regex !== null in toVdom()  #99 as @DAreRodz suggested.

It looks like they have modified `String.prototype.startsWith`. 

Screenshot 2022-11-29 at 18 04 46

That makes me think that the less we rely on native methods, the better 😄🤷‍♂️

We can fix it by doing [this change](https://github.com/WordPress/block-hydration-experiments/pull/99#issuecomment-1330956847).

DAreRodz:
> It looks like they have modified `String.prototype.startsWith`.

😳😳😳

michalczaplinski:
I've opened a PR to bundle the testing script with `webpack` instead of `esbuild`: 
- https://github.com/WordPress/block-hydration-experiments/pull/110

cbravobernal:
I did the script to get the posts with `1.csv` and I'm now running `2.csv`

SantosGuillamot:
I started retesting 4000 sites, and I encountered the same `hydrationError` in 8 of them. It seems that the script fails if there is a `ref` attribute in the HTML. I’ve opened an issue to discuss it with more information - https://github.com/WordPress/block-hydration-experiments/issues/116.

cbravobernal:
Running get posts on `3.csv`

SantosGuillamot:
We have tested all the sites and solved all the issues we encountered. Here's a summary and a recap of the process. Looking at the results, I believe we should feel confident enough for this initial phase.

### Results

Result | Count | % | Info
-- | -- | -- | --
Success | 145720 | 83.79 | Everything went fine
Cloudflare | 21165 | 12.17 | Skipped to avoid Cloudflare banning on our IPs
Timeout error | 3080 | 1.77 | Skipped because the request timed out
Error | 3025 | 1.75 | Skipped because the request didn’t return a valid response
403 error | 929 | 0.53 | Skipped because the request returned a 403 (forbidden)
Hydration error | 0 | 0 | The hydration algorithm failed
Mutations | 9 | 0 | The hydration algorithm produced DOM mutations
Total | 173919 | 100 | Total sites tested

### Main points

This is a list of the main issues we faced and the solutions proposed. We also iterated our testing script many times, but I’m not sharing anything related to that here.

* Remove comments: [Preact doesn’t support them during the hydration](https://href.li/?https://github.com/WordPress/block-hydration-experiments/issues/79) so we created this [Pull Request](https://href.li/?https://github.com/WordPress/block-hydration-experiments/pull/84) to remove them. We may need to add them back after the hydration if sites rely on them.
* [Nigloland.fr site issue](https://github.com/WordPress/block-hydration-experiments/pull/91#issuecomment-1315407994): This site is using a script that detects the removal of the comments with DOMSubtreeModified and makes extra mutations that are detected by our MutationObserver. We shouldn’t worry about it, but ideally we should preserve the comments.
* The less we rely on native methods, the better: [24life.ro modifies String.prototype.startsWith](https://href.li/?https://github.com/WordPress/block-hydration-experiments/pull/91#issuecomment-1330981396), which caused some mutations. As Luis suggests in the comment, the less we rely on native methods, the better. This was solved in [this pull request](https://href.li/?https://github.com/WordPress/block-hydration-experiments/pull/99#issuecomment-1330956847).
* [CDATA nodes](https://href.li/?https://github.com/WordPress/block-hydration-experiments/issues/109): We hadn’t taken into account the CDATA nodes ( nodeType 4), so they were returning some errors and registering mutations. As David explains [here](https://href.li/?https://github.com/WordPress/block-hydration-experiments/issues/109#issuecomment-1334073705), they are not allowed in HTML documents, but they are used inside SVGs. To solve it, we are converting them into Text nodes, implemented in [this pull request](https://href.li/?https://github.com/WordPress/block-hydration-experiments/pull/115/files).
* [xpacket nodes](https://href.li/?https://github.com/WordPress/block-hydration-experiments/issues/109): Similar to the previous point, we hadn’t yet considered the [ProcessingInstruction](https://href.li/?https://developer.mozilla.org/en-US/docs/Web/API/ProcessingInstruction) nodes (nodeType 7), and it was returning a hydration error. Again, it is not HTML, but it can be found in XML elements like SVGs. We are treating them like comments, implemented in [this pull request](https://href.li/?https://github.com/WordPress/block-hydration-experiments/pull/121).
* [ref attributes](https://href.li/?https://github.com/WordPress/block-hydration-experiments/issues/116): We also encountered that many sites were returning an error when there was a ref attribute in the content. At the end, we decided to [ignore them while generating virtual nodes](https://href.li/?https://github.com/WordPress/block-hydration-experiments/issues/116#issuecomment-1352854882).

SantosGuillamot:
As explained in [the issue](https://github.com/WordPress/block-hydration-experiments/issues/92), I believe we can consider this initial testing phase completed and successful, so I'm closing the PR as well.

PR #95, opened 2022-10-28
Use `page.evaluate` instead of `page.addScriptTag`

SantosGuillamot:
I was testing the workflow with different sites and in some of them ([TechCrunch](https://href.li/?https://techcrunch.com/) and [The Hill](https://href.li/?https://thehill.com/) for example), there was a Content Security Policy error:

`Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'unsafe-eval' 'none'`

I think that this can be solved using `evaluate` instead of `addScripTag`.

### DIFF
--- a/benchmark/index.mjs
+++ b/benchmark/index.mjs
@@ -1,3 +1,4 @@
+import * as fs from 'fs';
 import playwright from 'playwright';
 import { join } from 'path';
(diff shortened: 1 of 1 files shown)


michalczaplinski:
Great improvement!

PR #96, opened 2022-10-28
Don't add mutation if it is a comment removal

SantosGuillamot:
Right now, most of the mutations we are getting are comments, which is a use case we are already aware of. In order to keep the results as clean as possible, I believe we could skip those and don't add them to the database.

If we decide to merge this Pull Request, we may want to reset the `test_results.db` to keep it clean.


michalczaplinski:
Thanks @SantosGuillamot ! 

I will update the DB in the `perf-and-stress` testing branch

PR #98, opened 2022-10-28
Experiment: Add a custom proxy to automate Lighthouse testing

DAreRodz:
This experiment introduces a MITM proxy to intercept HTTP calls and inject our hydration library into the HTML response. The proxy is useful for automated testing with Lighthouse, which doesn't allow us to modify the HTML code of the page requested.

The library I used to build the proxy is https://github.com/joeferner/node-http-mitm-proxy. I think it's not a production-ready library, although it works well for this test. It's worth mentioning that the proxy server downgrades the performance a bit, but it could still be useful to get comparable data.

To start the proxy and the tests, you would run the following:

```bash
node benchmark/proxy.mjs
```

For the proxy to work with HTTPS, you must install the generated `.http-mitm-proxy/certs/ca.pem` certificate. To do that, you need to execute the following script:

```bash
npm run add-proxy-cert
```

Some next tasks could be:

- [ ] Store results in a database
- [ ] Make tests parallelizable
- [ ] Improve proxy performance


michalczaplinski:
Nice work @DAreRodz 🙂 

I wonder if we should stop pursuing the proxy approach for now though if it's also possible to inject the hydration script using Playwright as we're doing in https://github.com/WordPress/block-hydration-experiments/pull/91 

Is there any use case that would not be possible just Playwright? 🤔

luisherranz:
Hey @darerodz, are you still working on this?

DAreRodz:
@luisherranz, not really. And I would say there's no need to keep working on this (at least in the short term).

I've seen you've been closing issues and PRs lately, so if you also want to close this one, feel free. 🙂

luisherranz:
Ok. Thanks, David!

Issue #93, opened 2022-10-26
Limit the hydration of the Directives Hydration when the site doesn't support client-side navigations


luisherranz:
I've been reviewing the Directives Hydration approach with @mtias, and he suggested limiting the hydration to the smallest area possible when the site won't do client-side navigations.

Even though our initial tests suggest that the `toVdom` function is faster than we initially thought, I agree that we should try it because it will improve performance and increase backward compatibility. 

I've been thinking about it, and I believe it shouldn't add much complexity to the system. I've recorded a video to explain how I would approach it, which is basically:

- Use blocks as the "hydratable" unit (not directives or client components).
- Use `block.json` for the hydration opt-in.
- Blocks need to declare the type of hydration they require, depending on if the block interacts with their inner blocks or not (through context, error boundaries, suspense, etc.):
  - Isolated: 
    - If the block doesn't interact with its inner blocks 
    - The `toVdom` algorithm stops when it finds the `wp-inner-block` attributes.
  - Top-down: 
    - If the block interacts with its inner blocks (context, error boundaries, suspense, etc.)
    - The `toVdom` algorithm doesn't stop: everything in the tree below that block belongs to the same "app", including all their inner blocks. No interblock-sync is required (as opposed to the custom elements approach).

<a href="https://www.loom.com/share/3b4cb16c5b6c41d69f707b196cffb7f7">
    <img style="max-width:300px;" src="https://cdn.loom.com/sessions/thumbnails/3b4cb16c5b6c41d69f707b196cffb7f7-with-play.gif">
  </a>

https://www.loom.com/share/3b4cb16c5b6c41d69f707b196cffb7f7

https://excalidraw.com/#json=sk6HaigLP3YGxVq8tTo-H,HeYfhO2lmY4zC1uwdF5H6g

---

_I've also been exploring if there would be a way to "upgrade" a page that was hydrated using this partial vDOM hydration to directives hydration by reusing the vnodes created by the initial islands in the full app, but it requires manual mutation of the vnodes, and it feels to hacky for the moment. I also wasn't able to reuse the `useEffect` hooks (they are triggered again). [I mentioned it in Preact's Slack, and Marvin said they are exploring islands in Preact](https://preact.slack.com/archives/C3M9NTD16/p1666628614300359), so maybe we can collaborate with them in the future. For now, I would discard that path._

michalczaplinski:
[...]

PR #99, opened 2022-11-02
Replace `startsWith` in the `toVdom` function

michalczaplinski:
The result of `RegExp.prototype.exec()` can be null so we have to handle it.

PS. This was causing an error while running some of our stress-tests which is how I found it 🙂

### DIFF
--- a/src/runtime/vdom.js
+++ b/src/runtime/vdom.js
@@ -17,7 +17,9 @@ export default function toVdom(node) {
 			try {
 				val = JSON.parse(val);
 			} catch (e) {}
-			const [, prefix, suffix] = /wp-([^:]+):?(.*)$/.exec(name);
+			const result = /wp-([^:]+):?(.*)$/.exec(name);
+			if (!result) continue;
+			const [, prefix, suffix] = result;
 			wpDirectives[prefix] = wpDirectives[prefix] || {};
 			wpDirectives[prefix][suffix || 'default'] = val;
 		} else {
--- a/src/runtime/vdom.js
+++ b/src/runtime/vdom.js
@@ -10,19 +10,19 @@ export default function toVdom(node) {
 	if (node.nodeType === 3) return node.data;
 
 	if (attributes) {
-		for (let i = 0; i < attributes?.length; i++) {
-			const name = attributes[i].name;
-			if (name.startsWith('wp-')) {
+		for (let i = 0; i < attributes.length; i++) {
+			const n = attributes[i].name;
+			if (n[0] === 'w' && n[1] === 'p' && n[2] === '-' && n[3]) {
 				hasWpDirectives = true;
 				let val = attributes[i].value;
 				try {
 					val = JSON.parse(val);
 				} catch (e) {}
-				const [, prefix, suffix] = /wp-([^:]+):?(.*)$/.exec(name);
+				const [, prefix, suffix] = /wp-([^:]+):?(.*)$/.exec(n);
 				wpDirectives[prefix] = wpDirectives[prefix] || {};
 				wpDirectives[prefix][suffix || 'default'] = val;
 			} else {
-				props[name] = attributes[i].value;
+				props[n] = attributes[i].value;
 			}
 		}
 	}


michalczaplinski:
Hmm, I'm not sure what happened here - I can't reproduce this bug anymore and I also notice that the regex should never even fail because it's inside of `if (name.startsWith('wp-')) { }`. 

I think the only case it would fail is if there were an attribute like `wp-` (with nothing after the hyphen).

I ll have to investigate tomorrow because it's quite late now in my timezone 😅.

luisherranz:
`startsWith` should be [safe to use](https://caniuse.com/?search=startsWith) (it shouldn't be transpiled to a regular expression). That means this is probably the moment when we should start talking about the bundling target for the frontend hydration.

luisherranz:
To make this as close as possible to what we will ship in the future, I'd use Webpack to do the build and I'd start using [esmodules](https://babeljs.io/docs/en/options#targetsesmodules) in `babel/preset-env`.

michalczaplinski:
The `startsWith` was not transpiled in the built file when I was doing the testing so that was not the problem. I'm not sure what caused it in the end because I could not reproduce it in the end.

> I think the only case it would fail is if there were an attribute like wp- (with nothing after the hyphen).

What I meant here is that [this regex](https://github.com/WordPress/block-hydration-experiments/blob/2569df8f529604def4964a950b503eb8bc9828bf/src/runtime/vdom.js#L20) would fail if we had an an attribute like `wp-` (with nothing after the hyphen):

```js
const name = 'wp-'
/wp-([^:]+):?(.*)$/.exec(name); # returns null
```

> To make this as close as possible to what we will ship in the future, I'd use Webpack to do the build and I'd start using [esmodules](https://babeljs.io/docs/en/options#targetsesmodules) in babel/preset-env.

👍 Agreed on this one!

luisherranz:
Maybe we can replace this line:

```js
if (name.startsWith('wp-')) {
```

with

```js
if (name.startsWith('wp-') && name[3]) {
```

It should still be pretty fast.

SantosGuillamot:
FYI: It seems the site 24life.ro is facing this issue. It seems to be solved with the current status of this PR.

luisherranz:
I've tested these options:

```js
// 1
name.startsWith('wp-') && name[3]
// 2
name[0] === "w" && name[1] === "p" && name[2] === "-" && name[3]
// 3
const rg = /wp-\w/
rg.test(name);
```

Benchmark here: https://jsbench.me/9ilb2g41nb/1

3 is the slowest in every browser. 1 and 2 are equivalent in Chrome and Firefox, but 2 is twice as fast in Safari. Actually, looking at the results, it seems like `startsWith` is using `===` in Chrome and Firefox and a regular expression in Safari 😄

Unless someone else has another idea, let's use 2 for now.

luisherranz:
> FYI: It seems the site 24life.ro is facing this issue. It seems to be solved with the current status of this PR.

[It seems like they are replacing `String.prototype.startsWith`](https://github.com/WordPress/block-hydration-experiments/pull/91#issuecomment-1330981396).

luisherranz:
I've made the change. Ready for review 🙂

DAreRodz:
I'm fine adding these changes as it seems to be more performant. _However_, are we supposed to give this kind of support? If a site modifies `String.prototype.startsWith`, I would say that's their problem. 😅

michalczaplinski:
I can't approve it because it's my own PR.

I agree that we shouldn't have to worry about cases when native methods are modified but I approve anyway because it is faster in the end (thanks for pushing this forward @luisherranz! 🙂)

DAreRodz (approved):
Approved! ✅

SantosGuillamot (approved):
I've tested it, and it looks to be working fine. Thanks for the fix! 🙂

luisherranz:
>  If a site modifies String.prototype.startsWith, I would say that's their problem

> I agree that we shouldn't have to worry about cases when native methods

I agree with those statements, but I'd like to be extra careful with our `toVdom` to minimize backward incompatibilities. One thing is that you add a new block and it doesn't work because your `startsWith` is broken, and another thing is that you update WordPress and your site breaks. But yeah, it's not super important. Also, [if we limit the hydration](https://github.com/WordPress/block-hydration-experiments/issues/93) the DOM affected by `toVdom` that doesn't belong to interactive blocks will be way smaller.

PR #100, opened 2022-11-03
🎨 Server-side rendering of wp-show in PHP

luisherranz:
Co-authored with @santosguillamot.

Mario and I have created a simple proof of concept of how server-side rendering of a `wp-show` client component could look like. 

It uses a manual node script to parse the `render.php` file using [ultrahtml](https://github.com/natemoo-re/ultrahtml) (a simple HTML parser) and inject PHP functions. Then, we use those functions to evaluate the state and modify the markup accordingly.

We did this work on top of the `wpmoviesdemo` branch because we are using a block to test this out, and the `wpmoviesdemo` branch already had some blocks working fine. I didn't remove the `wpmoviesdemo` blocks, but I removed the Shared Elements transitions. _We haven't decided whether the WP Directives plugin should contain block examples or not._

This is very basic yet, and everything is hardcoded, but it demonstrates one possible approach. If we go ahead with this technique, we'd need to figure out good APIs to make this extensible.

<a href="https://www.loom.com/share/ce216cb0286e4b6ea421129814418c3e">
    
  </a>

https://www.loom.com/share/ce216cb0286e4b6ea421129814418c3e


luisherranz:
Closing this PR in favor of a possible run-time implementation based on the [WP_HTML_Tag_Processor](https://github.com/WordPress/gutenberg/pull/42485) that @ockham and @dmsnell have started investigating.

PR #101, opened 2022-11-15
Register 403 errors in the results.db

SantosGuillamot:
In some situations, for different reasons, we may not have access to the page, and we shouldn't register that as a success. For example, we are aware that Cloudflare could block it because we are doing too many requests. We are trying to solve Cloudflare use case, but there may be others, so I made this PR in case we want to register that.

### DIFF
--- a/benchmark/index.mjs
+++ b/benchmark/index.mjs
@@ -61,10 +61,26 @@ async function testUrl(url, browser) {
 		wordPressPage = await WordPressPage.create({ url });
 	}
 
+	let error403 = false;
+
 	try {
 		const page = await browser.newPage();
 
 		console.log(`Navigating to ${url}\n`);
+
+		page.on('response', async (msg) => {
+			const urlObject = new URL(msg.url());
+
+			// Check if the url returns a 403
+			if (
+				urlObject.hostname + urlObject.pathname === url &&
+				msg.status() === 403
+			) {
+				console.log('403 error');
+				error403 = true;
+				await page.close();
+			}
+		});
 
 		await page.goto(`http://${url}`, {
 			waitUntil: 'networkidle',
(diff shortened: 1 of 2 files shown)


cbravobernal (approved):

benchmark/index.mjs
```diff
@@ -66,11 +66,29 @@ async function testUrl(url, browser) {
...
+
 		await page.goto(`http://${url}`, {
 			waitUntil: 'networkidle',
 			timeout: 60000,
 		});
 
+		if (error403) {
+			throw '403 error';
+		}
+
```
DAreRodz: @SantosGuillamot, it would be simpler if you read the status from the `Response` object returned by `page.goto()`. Something like:

```suggestion
		const response = await page.goto(`http://${url}`, {
			waitUntil: 'networkidle',
			timeout: 60000,
		});

		if (response.status() === 403) {
			throw '403 error';
		}

```

DAreRodz: You can check the Response status docs here: https://playwright.dev/docs/api/class-response#response-status

SantosGuillamot: Much easier indeed 😄 Thanks for the suggestion @DAreRodz ! I've just commit it, although I'm using `response.status()` instead of `response.status`. Hope that's correct.

DAreRodz: Yes, that's correct; I've updated my suggestion. 😊

benchmark/index.mjs
```diff
@@ -66,11 +66,15 @@ async function testUrl(url, browser) {
...
 		console.log(`Navigating to ${url}\n`);
 
-		await page.goto(`http://${url}`, {
+		const response = await page.goto(`http://${url}`, {
 			waitUntil: 'networkidle',
 			timeout: 60000,
 		});
 
+		if (response.status() === 403) {
+			throw '403 error - ' + url;
```
DAreRodz: Hey, you are throwing a `string` here that is different than the one you are checking below!

`403 error - URL` will always be different than `403 error`

SantosGuillamot: Oh, I added that for testing and I forgot to delete it 😄 Changed! Thanks again 🙂

DAreRodz:
@SantosGuillamot feel free to merge this. 😊

PR #102, opened 2022-11-16
Check if the sites we are testing are using Cloudflare and skip them by default

SantosGuillamot:
I added some logic to check if the site tested is using Cloudflare and skip testing as it could be blocked by its bot detection system. Props to @luisherranz for the code 🙂 

Basically what I did was to add a new column in the `results.db` named `cloudflare`. It accepts the values `true` or `false`. Which are strings because I believe booleans are not supported. For the sites that are in Cloudflare and were already testing, I'm setting them to `null` because they could be false negatives if we hit the Cloudflare block page while testing them in the past.

Apart from that, I am reading a `cloudflare` variable from the CLI so, if at any moment we want to test Cloudflare's sites, just passing that argument should be enough.

There are many conditionals and I wasn't sure how to organize them in the best way. This is what I am checking:

* If we already checked it is a Cloudflare site but the CLI argument is not passed, skip the test.
* If we already checked it is (or not) a Cloudflare sire and it returned a `success`, skip the test.
* If we still don't know if it is a Cloudflare site, check it and update the database:
  * If it is a Cloudflare site, set `errorOrSuccess` to `null` because we could have encountered Cloudflare page and it could be a false negative.
  * If it is a Cloudflare site but the CLI argument is not passed, skip the test.
  * If it is NOT a Cloudflare site, and it was a success, skip the test.


cbravobernal (approved):

benchmark/index.mjs
```diff
@@ -52,15 +59,63 @@ async function testUrl(url, browser) {
...
+	// Skip test if it is a Cloudflare site and `testCloudflare` is not enabled
+	if (wordPressPage?.cloudflare === 'true' && !testCloudflare) {
+		console.log('Skip: Cloudfare site', url);
+		return;
+	}
+
+	// Skip test if we have already checked if it is a Cloudflare site and is a success
+	if (
+		wordPressPage?.cloudflare !== null &&
+		wordPressPage?.errorOrSuccess === 'success'
```
DAreRodz, line 75: Hey, what happens if `cloudflare` is `null` and `errorOrSuccess` is `success` at the same time? Shouldn't we skip the test as well? 🤔

SantosGuillamot: If `cloudflare` is `null` we can't assume it's okay even if it was `success`. As we were testing so many sites, we could have gotten blocked by Cloudflare and reached a blocking page that returned a `success`. We weren't contemplating that possibility so it could be a false negative.

I'm trying to handle that later in the code, [after checking if it is a Cloudflare site](https://github.com/WordPress/block-hydration-experiments/blob/cloudflare-check/benchmark/index.mjs#L95-L111).

DAreRodz: Okay, everything's good. Thanks, @SantosGuillamot, for the clarification. 😊

DAreRodz:
@SantosGuillamot feel free to merge this. 😊

PR #104, opened 2022-11-16
Avoid false positives from sites overwriting `Array.prototype.toJSON`

DAreRodz:
This PR fixes the case where some sites overwrite `Array.prototype.toJSON` with a function that serializes arrays as strings, as they yield false positives.

To prevent that, `toJSON` is deleted from the `Array` prototype right before serializing all the recorded mutations. That avoids serializing `addedNodes` and `removedNodes` as strings. The `toJSON` function is restored afterward to keep things the same.


luisherranz (approved):
LGTM!

PR #105, opened 2022-11-17
Add database name as parameter

cbravobernal:
Adding the database name as an optional parameter to the script may help prevent unwanted conflicts between database_results when executing the tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 build
+.DS_Store
--- a/benchmark/index.mjs
+++ b/benchmark/index.mjs
@@ -14,9 +14,9 @@ const {
 	_,
 	concurrency,
 	cloudflare: testCloudflare,
-	database = 'test_results',
+	database,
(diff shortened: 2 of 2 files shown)


SantosGuillamot (approved):
Looks great! 🙂 

I have a question not related to this PR. I see you used `database = 'test_results'`. Is it a way of defining the default value? Should we remove the [default object here](https://github.com/WordPress/block-hydration-experiments/blob/perf-stress-testing/add-name-db/benchmark/index.mjs#L19) and do the same for `cloudflare`?

cbravobernal:
> Is it a way of defining the default value?

Yes

> Should we remove the [default object here](https://github.com/WordPress/block-hydration-experiments/blob/perf-stress-testing/add-name-db/benchmark/index.mjs#L19) and do the same for cloudflare?

I didn't see that minimist has an option for setting defaults. I will try it and use it so we have the same way.

michalczaplinski (approved):
Looks good! Thanks @c4rl0sbr4v0!

PR #106, opened 2022-11-19
Fix registering of hydration errors

DAreRodz:
As I mentioned in https://github.com/WordPress/block-hydration-experiments/pull/91#issuecomment-1318990967, it looks like hydration errors were not registered correctly, and thus our hydration technique could have been failing in some sites without noticing. 😅 

This PR fixes that by returning the result of evaluating `__runHydration` directly from `page.evaluate`. Playwright automatically serializes that returned value, so there's no need to use `console.log` events for that.

Incidentally, this PR also removes the need to use [this hack](https://github.com/WordPress/block-hydration-experiments/pull/104) for those sites that overwrite `toJSON` functions for certain classes (e.g. `Array`).


michalczaplinski:
I wonder if this was not working already? I can see in the diff that we were catching the hydration error already and then console logging it (which was then captured by `page.on('console', () => {...})`.

I've tried running the updated script from this branch but I notice that many of the results are NULL now so that makes me think that we're forgetting to call the `wordPressPage.set('errorOrSuccess', '<value>');` somewhere...

DAreRodz:
> I wonder if this was not working already? I can see in the diff that we were catching the hydration error already and then console logging it (which was then captured by `page.on('console', () => {...})`.

Yup, that's what I thought! I think `page.on('console', ...)` is not reliable enough as the event listener may not be called.  At least it's what I experienced using the diff I shared in https://github.com/WordPress/block-hydration-experiments/pull/91#issuecomment-1318990967, which hardcodes a `throw` sentence. The exception was printed in the console, by the way, but the `console` callback didn't run.

> I've tried running the updated script from this branch but I notice that many of the results are NULL now so that makes me think that we're forgetting to call the wordPressPage.set('errorOrSuccess', '<value>'); somewhere...

This could happen if the site is a Cloudflare one. Can you check if the `cloudflare` value in the DB is `true` for those sites?

michalczaplinski:
Got it! Thanks David 🙂 

> This could happen if the site is a Cloudflare one. Can you check if the `cloudflare` value in the DB is `true` for those sites?

That is the case indeed! I think it would make sense to add this info to the `errorOrSuccess` column as well. Like `cloudflareError` or something like that.

michalczaplinski:
> Like cloudflareError or something like that.

Hope it's okay I took the liberty to add that now.

I've also moved the `wordPressPage.save();` to the end of the `try` clause in [86dfd4b](https://github.com/WordPress/block-hydration-experiments/pull/106/commits/86dfd4bbb486463e387ccad383141ccbbff2a2c4) because I think the test results might not have been added to the DB otherwise!

michalczaplinski (approved):
LGTM! 👍

SantosGuillamot:
> That is the case indeed! I think it would make sense to add this info to the errorOrSuccess column as well. Like cloudflareError or something like that.

I'm not sure about this one because we are actually **skipping** the test if the `--cloudflare` argument is not passed, is not an error. That's why I set it to null in [this PR](https://github.com/WordPress/block-hydration-experiments/pull/102). If we pass the argument, it will run the test and register if it is a success or an error.

DAreRodz:
@SantosGuillamot should we revert commit https://github.com/WordPress/block-hydration-experiments/pull/106/commits/756de0cf5919f8fd877ae6bf9aec5853c23526c4, then?

SantosGuillamot:
>@SantosGuillamot should we revert commit https://github.com/WordPress/block-hydration-experiments/commit/756de0cf5919f8fd877ae6bf9aec5853c23526c4, then?

Whatever you think is better. I thought that a non-tested site should remain as `NULL` but practically is the same to add something in the `errorOrSuccess`. As this is only for testing sites I wouldn't worry too much as far as it works.

SantosGuillamot (approved):

SantosGuillamot:
I've been talking with @DAreRodz, who agrees that registering it as a `cloudflareError` can be confusing because it isn't a Cloudflare error. To unblock the PR, I've reverted @michalczaplinski 's commit (sorry 🙏 ), and I'm merging it now.

michalczaplinski:
Makes sense @SantosGuillamot! I also think it doesn't make a big difference either way 🙂

PR #107, opened 2022-11-29
Make `attributes.length` optional

SantosGuillamot:
We have found a couple of sites where our hydration script is failing because `attributes.length` is `undefined`. It seems that adding `?` to it to make it optional solves the problem. These are the sites where it is failing:
* [tiens-donc.com](http://tiens-donc.com/)
* [yourcontentempire.com](http://yourcontentempire.com/)

### DIFF
--- a/src/runtime/vdom.js
+++ b/src/runtime/vdom.js
@@ -9,19 +9,21 @@ export default function toVdom(node) {
 
 	if (node.nodeType === 3) return node.data;
 
-	for (let i = 0; i < attributes?.length; i++) {
-		const name = attributes[i].name;
-		if (name.startsWith('wp-')) {
-			hasWpDirectives = true;
-			let val = attributes[i].value;
-			try {
-				val = JSON.parse(val);
-			} catch (e) {}
-			const [, prefix, suffix] = /wp-([^:]+):?(.*)$/.exec(name);
-			wpDirectives[prefix] = wpDirectives[prefix] || {};
-			wpDirectives[prefix][suffix || 'default'] = val;
-		} else {
-			props[name] = attributes[i].value;
+	if (attributes) {
+		for (let i = 0; i < attributes?.length; i++) {
+			const name = attributes[i].name;
+			if (name.startsWith('wp-')) {
+				hasWpDirectives = true;
+				let val = attributes[i].value;
+				try {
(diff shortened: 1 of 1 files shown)


src/runtime/vdom.js
```diff
@@ -9,7 +9,7 @@ export default function toVdom(node) {
 
 	if (node.nodeType === 3) return node.data;
 
-	for (let i = 0; i < attributes.length; i++) {
+	for (let i = 0; i < attributes?.length; i++) {
```
cbravobernal: `attributes[i].name` will also crash if attributes is undefined

SantosGuillamot: Oh, I thought that if `attributes` is undefined, it wasn't ever going to enter the `for` loop. What should we do then? Wrapping the whole `for` loop in a conditional like `if(attributes){ for (...){} }`?

cbravobernal: It is a curious case:

As you first in the file use destructure to get attributes (you could also define `let attributes;` which also defaults to undefined), in this case: `attributes?.length` is 'undefined'. 

So the comparison is `0 < undefined` which indeed is false 😆, so your thought is true, and it should not enter the loop.

So the code should work, but you may have unexpected comparisons 😅 . So I guess in this case preventing the loop with the conditional would be a good approach.

In case you try in a console without initializing attributes, it should crash 😄

SantosGuillamot: Thanks a lot for the explanation! It totally makes sense 🙂 

I've just pushed a new commit with your suggestion. Hope that's correct 😁

cbravobernal (approved):
LGTM!

luisherranz:
I've inspected why those nodes don't have `attributes`, and it's because they are CDATA (`nodeType = 4`). I'd prefer to make a conscious decision about what to do with CDATA, instead of just checking if attributes exist or not.

[I've opened an issue to discuss it](https://github.com/WordPress/block-hydration-experiments/issues/109).

SantosGuillamot:
Great! That totally makes sense. Should I revert the PR to ensure we are not skipping other use cases like CDATA?

luisherranz:
Let's see what comes up from https://github.com/WordPress/block-hydration-experiments/issues/109 first.

PR #110, opened 2022-11-30
Build the stress testing script with webpack instead of esbuild

michalczaplinski:
**Note**: This PR is to be merged into the `perf-and-stress-testing` branch

Update the build script for `benchmark/hydrationScript.js` to use webpack and remove `esbuild` from our dependencies.


cbravobernal (approved):
Works as expected! LGMT

PR #112, opened 2022-11-30
Script to get a post using Rest API to test it with the vDOM script.

cbravobernal:
I created an initial draft of a script that reads a CSV domains file, then, per domain it makes a request to the REST API in order to get the URL of the first post found and add it to another CSV.

I'm currently testing it and I'm having tons of 403 and sites with the REST API not enabled. Until I get a percentage to check if we should consider this approach, I will leave this PR as a draft.

There also some tasks pending or other possibilities:
  - [x] Check if we can use RSS to get the link of the post instead of the REST API.
  - [x] Use a DB to be able to stop and continue later with the script execution.

### DIFF
--- a/benchmark/getPostUris.mjs
+++ b/benchmark/getPostUris.mjs
@@ -1,0 +1,109 @@
+import * as fs from 'fs';
+import { parse } from 'csv-parse';
+import minimist from 'minimist';
+import { request } from '@playwright/test';
+
+// Get the CLI arguments for the file to use
+// and the number of pages to test concurrently.
+const { _, concurrency } = minimist(process.argv.slice(2), {
+	default: { concurrency: 10 },
+});
+const fileArg = _[0];
+if (typeof fileArg === 'undefined') {
+	console.error(
+		'\x1b[41m',
+		'\nPlease provide a CSV file as an argument\n\nFor example: npm run benchmark 1.csv\n'
+	);
+	process.exit(1);
+}
+
+// Run the benchmark
+(async () => {
+	try {
+		const domains = await getDomains();
+		// (B) WRITE TO FILE
+		await fs.writeFileSync('./benchmark/data/posts.csv', '');
+
+		await asyncParallelQueue(concurrency || 40, domains, async (url) => {
+			return await testUrl(url);
+		});
+	} catch (e) {
+		console.log(e);
+	}
+})();
+
+async function testUrl(url) {
+	try {
+		const context = await request.newContext({
+			baseURL: `https://${url}`,
+		});
+
+		try {
+			const response = await context.get(
+				'/wp-json/wp/v2/posts?post_per_page=1'
+			);
+			const href = await response.json();
+			await fs.appendFileSync(
+				'./benchmark/data/posts.csv',
+				`\n${href[0].link}`
+			);
+		} catch (error) {
+			console.log(error);
+		}
+	} catch (error) {
+		console.log(error);
+	}
+}
+
+async function getDomains() {
+	const domains = [];
+	const csvFile = new URL(`./data/${fileArg}`, import.meta.url).pathname;
+	return new Promise((resolve, reject) => {
+		fs.createReadStream(csvFile)
+			.pipe(
+				parse({
+					delimiter: ',',
+					from_line: 2,
+					skip_records_with_error: true,
+				})
+			)
+			.on('data', (row) => {
+				domains.push(row[0]);
+			})
+			.on('end', () => {
+				console.log('Domains Created');
+				resolve(domains);
+			})
+			.on('error', (error) => {
+				reject(error);
+			});
+	});
+}
+
+async function asyncParallelQueue(concurrency, items, func) {
+	const batch = items.slice(0, concurrency);
+	let promisesArray = batch
+		.map(func)
+		.map((p, i) => [i, p.then(() => i).catch(() => i)]);
+
+	// Create a "pool" of Promises.
+	const pool = new Map(promisesArray);
+
+	for (let index = 0; index < items.length; index++) {
+		const key = await Promise.race(pool.values());
+		pool.delete(key);
+		if (concurrency + index < items.length) {
+			pool.set(
+				concurrency + index,
+				func(items[concurrency + index])
+					.then(() => concurrency + index)
+					.catch(() => concurrency + index)
+			);
+		}
+	}
+}
+
+process.on('SIGINT', () => {
+	console.log('Caught interrupt signal');
+	process.exit(1);
+});
--- a/benchmark/getPostUris.mjs
+++ b/benchmark/getPostUris.mjs
@@ -3,11 +3,15 @@ import { parse } from 'csv-parse';
 import minimist from 'minimist';
 import { request } from '@playwright/test';
 import Parser from 'rss-parser';
+import { Sequelize } from 'sequelize';
+import { promises as dnsPromises } from 'dns';
+
+import { createModels } from './models.mjs';
 
 // Get the CLI arguments for the file to use
 // and the number of pages to test concurrently.
-const { _, concurrency, rss } = minimist(process.argv.slice(2), {
-	default: { concurrency: 10, rss: false },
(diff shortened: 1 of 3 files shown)


SantosGuillamot:
Thanks for opening this @c4rl0sbr4v0 !

Just to keep it in mind, some sites may have disabled access to the REST API and we will hit a "You are not allowed to access the REST API" (or something similar) message instead of the post.

benchmark/getPostUris.mjs
```diff
@@ -0,0 +1,109 @@
...
+
+async function testUrl(url) {
+	try {
+		const context = await request.newContext({
+			baseURL: `https://${url}`,
+		});
+
+		try {
+			const response = await context.get(
+				'/wp-json/wp/v2/posts?post_per_page=1'
```
luisherranz: Using `?rest_route=/wp/v2/posts&post_per_page=1` is a bit safer here.

cbravobernal:
> Just to keep it in mind, some sites may have disabled access to the REST API and we will hit a "You are not allowed to access the REST API" (or something similar) message instead of the post.

Yes! In that case, at the moment, we just skip it, catch and console. We can get the status and the response message if needed. It's the first draft just to keep the record and an initial number of URLs.

SantosGuillamot:
> Yes! In that case, at the moment, we just skip it, catch and console. We can get the status and the response message if needed. It's the first draft just to keep the record and an initial number of URLs.

Okay, great! 🙂 For reference, we could encounter at least two different messages:
* [REST API message](https://wpmoviesdemo.wpcomstaging.com/wp-json/wp/v2/templates): It returns a JSON saying it isn't allowed.
* [Warning page](https://wpcerber.com/wp-json/wp/v2/): It is a page saying you are not allowed.

benchmark/getPostUris.mjs
```diff
@@ -0,0 +1,109 @@
...
+		});
+	} catch (e) {
+		console.log(e);
+	}
+})();
+
+async function testUrl(url) {
+	try {
+		const context = await request.newContext({
+			baseURL: `https://${url}`,
```
SantosGuillamot: Not sure about the implications, but [in the current script we are using `http` instead of `https`](https://github.com/WordPress/block-hydration-experiments/blob/perf-and-stress-testing/benchmark/index.mjs#L125). Maybe it is because if we use `https` in a site without a certificate, the browser returns an error, and the other way they are usually redirected?

benchmark/getPostUris.mjs
cbravobernal: Done!

benchmark/getPostUris.mjs
```diff
@@ -0,0 +1,130 @@
...
+// Run the benchmark
+(async () => {
+	try {
+		const domains = await getDomains();
+		// (B) WRITE TO FILE
+		if (rss) {
+			await fs.writeFileSync('./benchmark/data/posts_rss.csv', '');
+		} else {
+			await fs.writeFileSync('./benchmark/data/posts.csv', '');
+		}
```
cbravobernal: This is not needed, as appendFileSync creates the file if it does not exist.

SantosGuillamot:
I've just thought that we may want to skip Cloudflare sites if we don't want to get blocked. We could do something like [this](https://github.com/WordPress/block-hydration-experiments/blob/perf-and-stress-testing/benchmark/index.mjs#L88-L93), I guess.

cbravobernal:
I think is stable enough to be review ready. I'm getting about a 50% of posts using this script.

SantosGuillamot:
I was testing it, and it looks great so far. I am encountering many sites that are registering a post with many encoded characters in the url. Something like this: `/%d8%a3%d9%87%d9%84%d8%a7-%d8%a8%d8%a7%d9%84%d8%b9%d8%a7%d9%84%d9%85/`.

Is that expected? More than 50% of the urls I got are like that.

benchmark/getPostUris.mjs
```diff
@@ -0,0 +1,183 @@
...
+
+import { createPostModel } from './models.mjs';
+
+// Get the CLI arguments for the file to use
+// and the number of pages to test concurrently.
+const {
+	_,
+	concurrency,
+	database,
+	cloudfare: testCloudflare,
```
SantosGuillamot: Small typo: It's `cloudflare` instead of `cloudfare`.

benchmark/getPostUris.mjs
```diff
@@ -0,0 +1,183 @@
...
+} = minimist(process.argv.slice(2), {
+	default: { concurrency: 10, database: 'posts_db', cloudfare: false },
```
SantosGuillamot: Small typo: It's `cloudflare` instead of `cloudfare`.

benchmark/getPostUris.mjs
```diff
@@ -0,0 +1,183 @@
...
+	}
+	if (!post) {
+		post = await Post.create({
+			url: url,
+		});
+	}
+	try {
+		// Skip test if it is a Cloudflare site and `testCloudflare` is not enabled
+		if (post?.cloudflare === 'true' && !testCloudflare) {
+			console.log('Skip: Cloudfare site', url);
```
SantosGuillamot: Small typo: It's `cloudflare` instead of `cloudfare`.

benchmark/getPostUris.mjs
```diff
@@ -0,0 +1,183 @@
...
+
+			// Remove record if it is a Cloudflare site
+			if (cloudflareSite) {
+				post.set('errorOrSuccess', null);
+			}
+			post.save();
+
+			// Skip test if `testCloudflare` is not enabled
+			if (cloudflareSite && !testCloudflare) {
+				console.log('Skip: Cloudfare site', url);
```
SantosGuillamot: Small typo: It's `cloudflare` instead of `cloudfare`.

benchmark/getPostUris.mjs
```diff
@@ -0,0 +1,183 @@
...
+	cloudfare: testCloudflare,
+} = minimist(process.argv.slice(2), {
+	default: { concurrency: 10, database: 'posts_db', cloudfare: false },
+});
+const parser = new Parser();
+const fileArg = _[0];
+if (typeof fileArg === 'undefined') {
+	console.error(
+		'\x1b[41m',
+		'\nPlease provide a CSV file as an argument\n\nFor example: npm run benchmark 1.csv\n'
```
SantosGuillamot: Just a small detail, but maybe we can change the error to say `npm run getposts` instead of `npm run benchmark`?

SantosGuillamot (changes requested):
I've added some minor comments. Apart from that, it looks great to me 🙂 

I was wondering as well if it could make sense to try fetching the post through the REST API when the RSS fails. Basically because I think you already had that logic. But maybe it is not worthy. I leave it up to you 🙂

benchmark/getPostUris.mjs
c4rl0sbr4v0: Copy-pasted, we need to update also the other file 😢

cbravobernal:
> I was wondering as well if it could make sense to try fetching the post through the REST API when the RSS fails.

We can do it, but also it will multiply the time of the script running by x1.8 more or less.  I don't know if it worths to be honest.

cbravobernal:
> I was testing it, and it looks great so far. I am encountering many sites that are registering a post with many encoded characters in the url. Something like this: /%d8%a3%d9%87%d9%84%d8%a7-%d8%a8%d8%a7%d9%84%d8%b9%d8%a7%d9%84%d9%85/.
Is that expected? More than 50% of the urls I got are like that.

Japanese and other special characters!!

zoes.tw/2021/10/23/1017%e6%a1%83%e5%9c%92amour%e8%ad%89%e5%a9%9a-%e5%87%b1%e9%88%9e%e6%9c%b1%e7%91%9c-%e5%a9%9a%e7%a6%ae%e9%8c%84%e5%bd%b1-%e7%b2%be%e8%8f%af%e7%89%88/ turns to be:
Screenshot 2022-12-02 at 11 16 13

SantosGuillamot (approved):
LGTM! 🙂

PR #114, opened 2022-12-01
Cloudflare stress testing

SantosGuillamot:
I'm creating this branch on top of https://github.com/WordPress/block-hydration-experiments/pull/91 to be able to test Cloudflare sites more safely. The idea is to test them in smaller batches, but I added a couple of things that should make it safer:
* I added the `https://github.com/berstend/puppeteer-extra/tree/master/packages/puppeteer-extra-plugin-stealth#readme` package, which claims to make things more difficult for Cloudflare to detect bots.
* For these tests, I removed the `asyncParallelQueue` because I assumed it is easier to bypass Cloudflare if we navigate the sites one by one. Moreover, if the batches are going to be smaller, this shouldn't be a big problem.

### DIFF
--- a/benchmark/index.mjs
+++ b/benchmark/index.mjs
@@ -1,6 +1,7 @@
 import * as fs from 'fs';
 import { parse } from 'csv-parse';
-import playwright from 'playwright';
+import playwright from 'playwright-extra';
+import StealthPlugin from 'puppeteer-extra-plugin-stealth';
 import { inspect } from 'util';
 import { Sequelize } from 'sequelize';
 import minimist from 'minimist';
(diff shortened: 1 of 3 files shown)


luisherranz:
@SantosGuillamot, are you still working on this?

SantosGuillamot:
I have a list of Cloudflare sites I'd like to test using this branch, but I haven't been able to do it yet. I still have it in mind, but I haven't prioritized it. I'll try to handle it in the coming weeks.

luisherranz:
Ok. Thanks for the update!

luisherranz:
Please, feel free to open a new discussion in the [Interactivity API category](https://github.com/WordPress/gutenberg/discussions/categories/interactivity-api) if you think this should be discussed further.

Issue #109, opened 2022-11-29
How to process CDATA in `toVdom`


luisherranz:
We've found some websites that have `CDATA` sections. One of them is https://www.tiens-donc.com/. I think @santosguillamot has more.

We need to decide what to do with those nodes. Maybe the first thing to investigate is what does Preact with them. Does it delete them? (as it does with the HTML comments)

I'm just opening the issue, so this is documented, but happy to see someone else take over on this 🙂

DAreRodz:
According to the [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/CDATASection):

> CDATA sections should not be used within HTML they are considered as comments and not displayed.

I would expect them to be comments (i.e., `nodeType = 8`), but you are right [here](https://github.com/WordPress/block-hydration-experiments/pull/107#issuecomment-1330915605): the CDATA sections have a different node type. 😓



> Maybe the first thing to investigate is what does Preact with them. Does it delete them? (as it does with the HTML comments)

I guess they don't even consider that case because you can't render CDATA sections with Preact. 🤷 In any case, I would treat them as comments if that's what browsers do.

SantosGuillamot:
>We've found some websites that have CDATA sections. One of them is https://www.tiens-donc.com/. I think @SantosGuillamot has more.

The only one I found apart from the one you mention is [yourcontentempire.com](https://href.li/?http://yourcontentempire.com/).

>Maybe the first thing to investigate is what does Preact with them. Does it delete them?

The only mention to CDATA in preact repo was [this Pull Request](https://github.com/preactjs/preact/pull/3337), which makes me think they treat CDATA the same way they treat comments. However, I don't know how to check that, I'm not familiar with their code.

SantosGuillamot:
I also found [this other issue ](https://github.com/facebook/react/issues/22518) in React repo talking about CDATA in SVGs.

I've been taking a look, and it seems it was a recommended way of adding style to SVGs (not sure if it still is). Moreover, if I navigate to one of the SVGs in [tiens-donc.com](https://www.tiens-donc.com/) (the Design grapiqhe logo at the footer) and remove the CDATA part, the style changes. On the other hand, pasting our script in the console, it removes the CDATA part, but it doesn't affect the style.

If this is something common, and we can affect the layout of the page, I wouldn't feel comfortable just ignoring them.

DAreRodz:
[...]

PR #115, opened 2022-12-02
Handle CDATA nodes

SantosGuillamot:
I'm creating this Pull Request with the solution proposed at https://github.com/WordPress/block-hydration-experiments/issues/109

I also reverted https://github.com/WordPress/block-hydration-experiments/pull/107. I made that one trying to solve the problem with `attributes` being `undefined` caused by the `CDATA` nodes. However, now that this is handled properly, it makes sense to revert that change in case we encounter another similar use case with undefined attributes. We should identify them and not skip them.

### DIFF
--- a/src/runtime/vdom.js
+++ b/src/runtime/vdom.js
@@ -8,6 +8,10 @@ export default function toVdom(node) {
 	let hasWpDirectives = false;
 
 	if (node.nodeType === 3) return node.data;
+	if (node.nodeType === 4) {
+		node.replaceWith(new Text(node.nodeValue));
+		return node.nodeValue;
+	}
 
 	for (let i = 0; i < attributes.length; i++) {
 		const n = attributes[i].name;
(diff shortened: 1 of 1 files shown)


DAreRodz (approved):
Thanks, @SantosGuillamot! I guess we'll have to sync the `stress-testing` branch (https://github.com/WordPress/block-hydration-experiments/pull/91) with these changes.

Can you also leave a note in the [Tracking issue: WordPress Directives Plugin 🎨 #80](https://github.com/WordPress/block-hydration-experiments/issues/80)? 😊

SantosGuillamot:
>Thanks, @SantosGuillamot! I guess we'll have to sync the stress-testing branch (https://github.com/WordPress/block-hydration-experiments/pull/91) with these changes.

Yes, I was planning to do so once this one is merged. I'll try to do it now.

>Can you also leave a note in the https://github.com/WordPress/block-hydration-experiments/issues/80? 😊

Just did it. Thanks!

PR #118, opened 2022-12-13
Experiment with `WP_HTML_Tag_Processor`

ockham:
# Superseded by #125.

~Early-stage experiments, mostly in order to discover our requirements from `WP_HTML_Tag_Processor`.~

~**Doesn't work yet!**~

~PR based on #100 -- I've taken the example blocks from there and removed the extra build step that was introduced there. Setting the base branch to #100's for now so it's clearer what I've changed.~


ockham:
Here's a preliminary TODO list, based on my inline comments in the code so far. I have some questions (in italics); if you have any answers (or opinions), please comment below! 😊 

- [ ] Allow wrapping a tag's inner contents. E.g. `<wp-show when="context.myblock.open">abcd</wp-show>` becomes `<wp-show when="context.myblock.open"><template>abcd</template></wp-show>` (if `context.myblock.open` evaluates to `false`).
  - It seems like https://github.com/WordPress/gutenberg/pull/46080 will support replacing tags.
    - _Should I review this, are there any major changes planned, or are y'all thinking of a totally different way of doing this?_ cc/ @dmsnell @adamziel 
    - _Would it make sense to add a dedicated method for wrapping a tag's inner contents?_
- [ ] Properly parse expressions in attributes like `when`.
  - Includes nested JS-style objects (e.g. `context.myblock.open`).
  - _Basic logic operations (negation, and, or)? Comparison?_ cc/ @WordPress/frontend-dx 
- [ ] Allow replacing a tag that has a directive attached.
  - _What would the API for something like that look like? Assuming that we allow registration of custom directives -- do the directive "processors" get access to the tag processor?_
- [ ] Add some directive/components registration mechanism.
  - Maybe similar to blocks, patterns, etc? `register_directive`, takes a directory with a `directive.json` which includes information such as `attributes`, `render`, context, ...? 
- [ ] _How do the APIs of signatures relate to their counterpart components? Can there be a 1:1 mapping (i.e. an adapter)?_
- [ ] Allow getting attributes that match a certain pattern. E.g. `wp-bind` [actually looks something like](https://github.com/WordPress/block-hydration-experiments/blob/403a19576309d89bb86fe776c7a0bb599cc3b7f3/src/blocks/post-favorite/render.php#L13) `` in practice (to bind the value of the image `src` attribute to the expression). This means we need syntax like `getAttribute( 'wp-bind:*' )`.

ockham:
> _Would it make sense to add a dedicated method for wrapping a tag's inner contents?_

I forgot that @dmsnell had shared with me a possible syntax for this:

```php
$content = tags->get_content_inside_balanced_tags();
$tags->set_content_inside_balanced_tags(
    "<template>{$content}</template>"
);
```

This uses `get_content_inside_balanced_tags`, which Dennis has already implemented in https://github.com/WordPress/gutenberg/pull/46345 🎉

wp-directives.php
```diff
@@ -176,31 +152,73 @@ function wpx($data)
...
+	$directives = array(
+		'wp-context' => 'process_wp_context',
+		'wp-show' => 'process_wp_show',
+	);
+
+	$tags = new WP_HTML_Tag_Processor( $block_content );
+
+	$context = array();
+	while ( $tags->next_tag() ) {
+		$tag_name = strtolower( $tags->get_tag() );
```
dmsnell, line 176: we chose to use upper case in line with HTML specs. it might be less familiar but more appropriate to simply check `'WP-SHOW' === $tags->get_tag()` 🤷‍♂️ 

that is, embrace the language of the system instead of the convention of the community. it avoids an additional string conversion if that matters.

there's also `strcasecmp()` if prefer not to think about it

wp-directives.php
```diff
@@ -176,31 +152,73 @@ function wpx($data)
...
+		if ( array_key_exists( $tag_name, $directives ) ) {
+			$attributes = array();
+			// TODO: We need some sort of allowlist of attributes per component.
+			// Maybe in a directive.json?
+			if ( 'wp-show' === $tag_name ) {
+				$attributes['when'] = $tags->get_attribute( 'when' );
+			}
+			call_user_func_array( $directives[$tag_name], array( $attributes, &$context ) );
```
dmsnell: the way I've been exploring this is by passing the `WP_HTML_Tag_Processor` around to functions, so here that might be `call_user_func_array( $directives[ $tag_name ], array( $tags, $attributes, &$context ) )`

of course, the Tag Processor will not be able to modify the markup. eventually this might need to be created as the `WP_HTML_Processor` instead, or some other thing. I've toyed with two ideas:

 - `$tags = new WP_HTML_Processor( $html )` instead of creating the tag processor
 - `new WP_HTML_Processor( $tags )` to wrap the tag processor

in the first method it suggests we might see people stop ever invoking `WP_HTML_Tag_processor()` directly, but avoids problems with encapsulation of the tag processor.

in the second method there's an extra step one has to take before walking into _the danger zone_ where performance and reliability drop off, but it allows for a stronger separation of those zones.

wp-directives.php
ockham: > we chose to use upper case in line with HTML specs.

I thought so! I'm mostly converting to lowercase because of the check on [L174](https://github.com/WordPress/block-hydration-experiments/pull/118/files#diff-444eef806a52fae63fdfc7790460035223771b962924b10826f34d71c7e2e09dR174) where we look up the directive/component name in a [whitelist](https://github.com/WordPress/block-hydration-experiments/pull/118/files#diff-444eef806a52fae63fdfc7790460035223771b962924b10826f34d71c7e2e09dR164-R167). We'll have something like that in the long run, alongside with some registration mechanism where users can register their own directives/components; and I'd rather use lowercase as the (community) convention there 😬 

> there's also strcasecmp() if prefer not to think about it

That sounds like a good option, thank you for suggesting 👍

dmsnell: I've considered adding `is_tag()` to the tag processor to avoid this, but so far haven't because it hasn't warranted itself enough of a problem.

this is less about the tag processor, but since it's related, did you consider a `switch` vs. the array? seems like that could remove some indirection and make the control flow more obvious and searchable

```php
$supported_directives = array(
	'WP-CONTEXT',
	'WP-SHOW',
);

while ( $tags->next_tag() ) {
	$directives = array();

	// Find the directives.
	foreach ( $supported_directives as $directive ) {
		// We could have an HTML directive like `<wp-show>`
		if ( $directive === $tags->get_tag() ) {
			$directives[] = array( $directive, 'tag' );
		}

		// We could have multiple HTML attribute directives like `<div wp-show=…>`
		$attribute_content = $tags->get_attribute( $directive );
		if ( null !== $attribute_content ) {
			$directives[] = array( $directive, 'attribute' );
		}
	}

	// Process the found directives.
	foreach ( $directives as $found_directive ) {
		list( $directive, $type ) = $found_directive;

		switch ( $directive ) {
			case 'WP-CONTEXT':
				$attributes = $type === 'tag'
					? array()
					: $tags->get_attribute( 'wp-context' )
				process_wp_context( $attributes, &$context );
				continue;

			case 'WP-SHOW':
				$attributes = $type === 'tag'
					? array( 'when' => $tags->get_attribute( 'when' ) )
					: $tags->get_attribute( 'wp-show' )
				process_wp_show( $attributes, &$context );
				continue;

			// Everything else is unsupported
		}
	}
}
```

I know this is premature feedback, but in the past I have also found this kind of control-flow confusing, where there's no obvious way to find in the code the functions that are called or where they are called. Ignore as you wish, just food for thought.

ockham: I think the `switch` would make a ton of sense if we had a hard-wired list of directives. However, since we're planning to allow users to register their own, I'm inclined to leave the code as-is, as the current approach is a bit closer to that already 😊

wp-directives.php
ockham: Yeah, IMO, this is one of the more complex (and interesting) problems here 😅 

Given the two options you're suggesting, I'm currently leaning towards the second one (because of danger zone separation).

I'm a bit concerned with the potential leakage of giving each directive processor access to the "entire" Tag Processor. Ideally, I'd love if a directive processor could only operate on its "own" (outer) HTML. I'm not sure how feasible that would be, since we'd still have to "stitch" the modified HTML back into the surrounding parts 🤔

wp-directives.php
```diff
@@ -176,31 +152,73 @@ function wpx($data)
...
+
+	$tags = new WP_HTML_Tag_Processor( $block_content );
+
+	$context = array();
+	while ( $tags->next_tag() ) {
+		$tag_name = strtolower( $tags->get_tag() );
+		if ( array_key_exists( $tag_name, $directives ) ) {
+			$attributes = array();
+			// TODO: We need some sort of allowlist of attributes per component.
+			// Maybe in a directive.json?
```
luisherranz: Ideally, we should pass all the attributes to the component, so there should be no need for a whitelist.

I guess you suggest this because there's no way to do `$tags->get_all_attributes();` as of today, or is there any other reason?

wp-directives.php
```diff
@@ -176,31 +152,73 @@ function wpx($data)
...
-{
-	$array = [];
-	foreach ($prop_entries as list($key, $value)) {
-		$array[$key] = $value;
-	}
-	return $array;
+function process_wp_context( $directive_content, &$context ) {
+	$new_context = json_decode( $directive_content, true );
+	// TODO: Error handling.
+	$context = array_replace_recursive( $context, $new_context );
```
luisherranz: Good catch. I was using `array_merge_recursive` before, but the correct function is `array_replace_recursive`.

wp-directives.php
```diff
@@ -176,31 +152,73 @@ function wpx($data)
 	$wpx = array_merge_recursive($wpx, $data);
```
luisherranz: We can replace it with `array_replace_recursive` here as well.

```suggestion
	$wpx = array_replace_recursive($wpx, $data);
```

luisherranz (commented):
This is exciting 🙂👏

wp-directives.php
ockham: Yeah, I had `array_merge_recursive` first myself but noticed it didn't do what I wanted it to 😅

wp-directives.php
ockham: The absence of a `get_all_attributes()` might've prompted this, but I was also genuinely thinking that an allowlist might make sense -- in order to limit the information the directive processor was allowed to see, and to make those attributes more explicit (akin to block attribute definitions) 🤔

But maybe that's a moot point, since we might have to give the directive processor access to the Tag Processor anyway ([see](https://github.com/WordPress/block-hydration-experiments/pull/118#discussion_r1047716197)).

ockham:
@luisherranz Thank you for your feedback! While you're here, can I ask your opinion about this question:

> _How do the APIs of signatures relate to their counterpart components? Can there be a 1:1 mapping (i.e. an adapter)?_

E.g. for context, we have for the directive:

```html
<div wp-context='{ "myblock": { "open": false } }'>
```

and for the component:

```html
<wp-context data='{ "myblock": { "open": false } }'>
```

Can we always assume that the component will have one named attribute that maps to what would be the value of the corresponding directive? (E.g. for `wp-context`, that would be `data` in the example above.) (You might have a more informed opinion through your familiarity with Alpine than I do 😊 )

In that case, we probably need some way to set that attribute's name in the directive declaration.

ockham:
@dmsnell I added another item to my [comment](https://github.com/WordPress/block-hydration-experiments/pull/118#issuecomment-1349095527) above:

> Allow getting attributes that match a certain pattern. E.g. `wp-bind` [actually looks something like](https://github.com/WordPress/block-hydration-experiments/blob/403a19576309d89bb86fe776c7a0bb599cc3b7f3/src/blocks/post-favorite/render.php#L13) `` in practice (to bind the value of the image `src` attribute to the expression). This means we need syntax like `getAttribute( 'wp-bind:*' )`.

Curious to hear your thoughts. Maybe the alternative would be a `get_all_attributes` method, as Luis [suggested](https://github.com/WordPress/block-hydration-experiments/pull/118#discussion_r1048493137)?

ockham:
I've updated the code to:
- Use the latest `WP_HTML_Tag_Processor` -- and `WP_HTML_Processor` -- from https://github.com/WordPress/gutenberg/pull/46345.
- Unify the directive processors so they can be used for both (attribute) directives and components ("HTML directives").
- Changed the directive processors to accept their inner HTML content (as obtained by `get_content_inside_balanced_tags`) as their first argument, and return the (potentially modified) content.

What's still missing is a method to _write_ the modified content back to `WP_HTML_Processor` -- see https://github.com/WordPress/block-hydration-experiments/pull/118#issuecomment-1349184703, or https://github.com/WordPress/block-hydration-experiments/pull/118#discussion_r1047716197.

wp-directives.php
```diff
@@ -142,65 +144,104 @@ function wp_directives_client_site_transitions_option()
...
+	while ( $tags->next_tag() ) {
+		$tag_name = strtolower( $tags->get_tag() );
+		if ( array_key_exists( $tag_name, $directives ) ) {
+			$directive_content = $tags->get_attribute( $directives[$tag_name]['default_attribute'] );
+
+			$bookmark = "{$tag_name}_start";
+			$tags->set_bookmark( $bookmark );
+			$content = $tags->get_content_inside_balanced_tags();
+			$tags->seek( $bookmark );
+			$tags->release_bookmark( $bookmark );
```
ockham: This is a bit verbose -- especially considering that `get_content_inside_balanced_tags` actually saves bookmarks that delimit the inner content.

ockham: Left a related note here: https://github.com/WordPress/gutenberg/pull/46345#issuecomment-1351779071

wp-directives.php
dmsnell: > I'm not sure how feasible that would be, since we'd still have to "stitch" the modified HTML back into the surrounding parts 🤔

I think this will end up being one of those fundamental limitations of the system. What we have is HTML, whereas what we don't have is DOM.

So while what you are talking about is possible, it's not currently possible inside the tag processor, and that may never change.

```php
foreach ( … ) {
	$inner_html = $tags->get_content_inside_balanced_tags();
	$inner_tags = new WP_HTML_Tag_Processor( $inner_html );

	$replacement_html = process_directive( …, $inner_tags );
	$tags->set_content_outside_balanced_tags( $replacement_html );
}
```

wp-directives.php
dmsnell: Everything in the linked PR is still preliminary and all names and classes might change drastically; not moving the pointer is reasonable and I'll add that in the branch.

wp-directives.php
dmsnell: having an explicit list of attribute is definitely better for the tag processor. HTML parsing is costly, and the tag processor is focused on low-level performance. while it's possible to open up `get_all_attributes()` I'd want to be very sure we need that, because we're working hard to ensure that whatever API we expose is fully warranted (since we expect people to use whatever is exposed).

granted, calling `get_all_attributes()` only incurs a performance penally and memory allocations when you ask it to, but it might not be obvious that you could end up ballooning the entire render process when a given tag has a lot of attributes, or when you process every tag in the document, as is done here, or when a number of tags have lengthy data-URI attributes.

if we _are able_ to do this without `get_all_attributes()` I'd prefer we explore that first. conveniences can always be added later but unwanted additions will be basically impossible to get rid of. in fact, needing to have an explicit list of attributes seems like a good thing in general here, because why would we want to parse and collect everything if we only need one or two? think of all the `style` attributes and potential data-URIs and lengthy `class` attributes…this can quickly lead to many MBs of memory overhead, not to mention substantially slow down processing and become the obstacle in the render path.

wp-directives.php
dmsnell: on another note, given that as written any HTML tag can involve multiple directives, and given that everything can be specified within the directive attributes, what's the point of having the directive tags? why not solely rely on the attributes?

wp-directives.php
luisherranz: > But maybe that's a moot point, since we might have to give the directive processor access to the Tag Processor anyway

Yes, that's probably going to be the case so let's keep this going on to see where it ends 🙂

luisherranz:
These are my opinions about your questions 🙂

---

> Basic logic operations (negation, and, or)? Comparison?

It is true that PHP and JavaScript have the same syntax for most logical operations and building a small parser sounds feasible, but I would try to avoid it, at least at first.

The one we could include initially is negation (`!`) which is the only operator that doesn't need two variables and therefore the simplest. It will also be the most common since there is no directive for `else`:

```html
<wp-show when="context.open">
  <span>I'm open</span>
</wp-show>
<wp-show when="!context.open">
  <span>I'm closed</span>
</wp-show>
```

> Allow replacing a tag that has a directive attached

I'm not sure I understand this question. Do you mean something like this?

```html
<div wp-change-tag="span"></div>
<!-- Result: -->
<span wp-change-tag="span"></span>
```

> Maybe similar to blocks, patterns, etc? register_directive, takes a directory with a directive.json which includes information such as attributes, render, context, ...?

That's actually a great idea! Although I guess it depends on what information we need. If it's just a name, a PHP function and a JS file, it may be a bit overkill to have to declare that in a JSON file 🤔

What other information do you think will be needed?

> How do the APIs of signatures relate to their counterpart components? Can there be a 1:1 mapping (i.e. an adapter)? [...] Can we always assume that the component will have one named attribute that maps to what would be the value of the corresponding directive?

I think that is more of a question about promoting some type of naming convention/standardization than forcing a strict rule.

I've actually been thinking lately about the fact that components can read multiple values (one per attribute) and directives only one (the value of its own attribute) and what syntax we should promote to standardize multi-value directives to keep the ecosystem consistent.

Imagine a directive/component that needs a handler and has optional options. That's straightforward in a component:

```html
<wp-swipe handler="myblock.swipeHandler" options="myblock.swipeOptions">
  ...
</wp-swipe>
```

But not so in the directive:

```html
<div wp-swipe="myblock.swipeHandler" options??>...</div>
```

There are several syntaxes we could promote, but I'm not convinced about any of them. We can open a separate issue to discuss it, although it's not urgent as it'd be only for standardization and not part of the API per se.

```html
<div
  wp-swipe="myblock.swipeHandler"
  wp-swipe-options="myblock.swipeOptions"
  wp-swipe:options="myblock.swipeOptions"
  wp-options:swipe="myblock.swipeOptions"
  wp-swipe="myblock.swipeHandler, myblock.swipeOptions"
  wp-swipe='{ "handler": "myblock.swipeHandler", "options": "myblock.swipeOptions" }'
  wp-swipe.option1.value1.option2.value2="myblock.swipeHandler"
>
  ...
</div>
```

luisherranz:
> This means we need syntax like `getAttribute( 'wp-bind:*' )`

Oh, that's true. We need something like that. Although there can be multiple of them in a tag and we need all of them.

By the way, the way this is currently coded in JavaScript is:

```js
const wpDirectives = {};
for (let i = 0; i < attributes.length; i++) {
  if (attributes[i].name.startsWith("wp-")) {
    const [, prefix, suffix] = /wp-([^:]+):?(.*)$/.exec(attributes[i].name);
    wpDirectives[prefix] = wpDirectives[prefix] || {};
    wpDirectives[prefix][suffix || "default"] = attributes[i].value;
  }
}
```

All the directive attributes end up in an object with this shape:

```html
<div
  wp-effect="myblock.someEffect"
  wp-effect:other-stuff="myblock.otherEffect"
  wp-effect:yet-another="myblock.yetAnotherEffect"
  wp-class:open="myblock.toggleOpen"
  wp-context='{"open": false}'
>
  ...
</div>
```

```js
{
  wp: {
    effect: {
      default: "myblock.someEffect",
      ["other-stuff"]: "myblock.otherEffect",
      ["yet-another"]: "myblock.yetAnotherEffect",
    },
    class: {
      open: "myblock.toggleOpen"
    },
    context: {
      default: '{"open": false}'
    }
  }
}
```

We don't need to extract the attributes for them if they have access to a Tag Processor instance, but they need to be able to do it themselves with a method.

wp-directives.php
luisherranz: > given that as written any HTML tag can involve multiple directives, what's the point of having the directive tags?

In some cases, their semantics could be preferable as these act more like web components. We haven't included any in the [preliminary roadmap](https://frontity.wordpress.com/2022/11/29/interactivity-api-preliminary-technical-roadmap/), but it'd be nice to ensure that the system can support them.

---

By the way, I like the distinction between _directive attributes_ and _directive tags_. We've been using components so far, but it's a bit vague 🙂

wp-directives.php
ockham: > So while what you are talking about is possible, it's not currently possible inside the tag processor, and that may never change.

Maybe we could do a variant of the code you shared with a method that directly returns a `WP_HTML_Tag_Processor` (or `WP_HTML_Processor`) which will include all bookmarks that apply to it and which can then be used to modify it in a number of defined ways before stitching it back in?

```php
foreach ( … ) {
	$inner_tags = $tags->get_tags_inside_balanced_tags(); // Probably needs to be WP_HTML_Processor.

	$process_directive( …, $inner_tags );
	$tags->set_tags_inside_balanced_tags( $inner_tags );
}
```

Among the limited ways of modifying `WP_HTML_Processor` could be something like "wrap in tag". This would cover our need to wrap things in `<template>`s.

Would a mechanism like that be able to preserve the integrity of the "outer" `WP_HTML_Processor`? Am I missing something?

---

(If we limit the ways of modifying that strongly, it's arguable that we could simply introduce a shorthand method to `wrap_content_inside_balanced_tags()`.)

ockham:
I've filed #120 to capture the current state of the TODO list. I plan to break out items into individual issues so we can discuss them more in-depth.

ockham:
Thank you very much for your replies @luisherranz! They'll help inform the next steps we'll take 😊 

Some follow-up questions:

Do we even expect a lot of directives to have more than one attribute? IMO, a syntax like `wp-bind:src` makes a lot of sense since it makes clear that it's referencing something else -- a `src` attribute. However, if we expect some directives to receive a _lot_ of attributes, things might get a bit messy anyway. Do we know how e.g. Alpine is dealing with this? Are there any directives that accept multiple attributes?

Generally, do we have a sense of what transformations we need to support? E.g. I know we want to wrap a `<wp-show>`'s children in `<template>`; but what else do we need? The item "Allow replacing a tag that has a directive attached" on my TODO list was related to that. I meant indeed something like

```html
<div wp-change-tag="span"></div>
<!-- Result: -->
<span wp-change-tag="span"></span>
```

Although of course a bit more implicit. Something like

```html
<div wp-do-something-based-on-criterion="context.some.information"></div>
<!-- Result: -->
<something-really-different maybe-with-some="attribute">abcd</something-really-different>
```

luisherranz:
This is the list of currently known markup modifications required by the different directives:

- **wp-show**

  Wrap its children with a template.

  ```html
  <!-- Written by the developer -->
  <wp-show when="myblock.falsyValue">Content...</wp-show>

  <!-- After processing -->
  <wp-show when="myblock.falsyValue">
    <template>Content...</template>
  </wp-show>
  ```

  ```html
  <!-- Written by the developer -->
  <div wp-show="myblock.falsyValue">Content...</div>

  <!-- After processing -->
  <div wp-show="myblock.falsyValue">
    <template>Content...</template>
  </div>
  ```

- **wp-text, wp-html**

  Replace children.

  ```html
  <!-- Written by the developer -->
  <wp-text data="myblock.someValue"></wp-text>

  <wp-html data="myblock.someValue"></wp-html>

  <!-- After processing -->
  <wp-text data="myblock.someValue">
    This is the value stored in `myblock.someValue`.
  </wp-text>

  <wp-html data="myblock.someValue">
    This is the value stored in `myblock.someValue`. It can contain
    <strong>HTML</strong>.
  </wp-html>
  ```

  ```html
  <!-- Written by the developer -->
  <span wp-text="myblock.someValue"></span>

  <span wp-html="myblock.someValue"></span>

  <!-- After processing -->
  <span wp-text="myblock.someValue">
    This is the value stored in `myblock.someValue`.
  </span>

  <span wp-html="myblock.someValue">
    This is the value stored in `myblock.someValue`. It can contain
    <strong>HTML</strong>.
  </span>
  ```

- **wp-bind, wp-class, wp-style**

  Replace attributes.

  ```html
  <!-- Written by the developer -->
  <img wp-bind:src="myblock.imageSource" />
  <div wp-class:red="myblock.someValue" class="blue"></div>
  <div wp-style:color="myblock.someValue" style="background: blue;"></div>

  <!-- After processing -->
  <img wp-bind:src="myblock.imageSource" src="Value of myblock.imageSource" />
  <div wp-class:red="myblock.truthyValue" class="blue red"></div>
  <div
    wp-style:color="myblock.greenValue"
    style="background: blue; color: green;"
  ></div>
  ```

- **wp-each**

  Iterate over an array:

  ```html
  <!-- Written by the developer -->
  <wp-each data="myblock.someArray" key="item.id">
    <span wp-text="item.name"></span>
    <span wp-text="item.surname"></span>
  </wp-each>

  <!-- After processing -->
  <wp-each data="myblock.someArray" key="item.id">
    <template>
      <span wp-text="item.name"></span>
      <span wp-text="item.surname"></span>
    </template>
    <span wp-text="item.name">John</span>
    <span wp-text="item.surname">Snow</span>
    <span wp-text="item.name">Ned</span>
    <span wp-text="item.surname">Stark</span>
    <span wp-text="item.name">Jaime</span>
    <span wp-text="item.surname">Lannister</span>
  </wp-each>
  ```

  ```html
  <!-- Written by the developer -->
  <ul wp-each="myblock.someArray" wp-each:key="item.id">
    <li>
      <span wp-text="item.name"></span>
      <span wp-text="item.surname"></span>
    </li>
  </ul>

  <!-- After processing -->
  <ul wp-each="myblock.someArray" wp-each:key="item.id">
    <template>
      <li>
        <span wp-text="item.name"></span>
        <span wp-text="item.surname"></span>
      </li>
    </template>
    <li>
      <span wp-text="item.name">John</span>
      <span wp-text="item.surname">Snow</span>
    </li>
    <li>
      <span wp-text="item.name">Ned</span>
      <span wp-text="item.surname">Stark</span>
    </li>
    <li>
      <span wp-text="item.name">Jaime</span>
      <span wp-text="item.surname">Lannister</span>
    </li>
  </ul>
  ```

  _The `template` may not be strictly necessary if we enforce a single root element as [Alpine does](https://alpinejs.dev/directives/for). Similar to [sidewind](https://sidewind.js.org/ssr/)._

- **wp-slot and wp-fill**

  Move markup from one place to another.

  _Probably out of the scope of the first version, but it'd be great to have an idea of how we can do this._

  ```html
  <!-- Written by the developer -->
  <wp-slot name="answer">
    <!-- default content if no fills exist -->
    No answer
  </wp-slot>

  <!-- somewhere else in the document -->
  <wp-fill slot="answer">
    <span>I'm the first answer!</span>
  </wp-fill>
  <wp-fill slot="answer">
    <span>I'm the second answer!</span>
  </wp-fill>

  <!-- After processing -->
  <wp-slot name="answer">
    <wp-fill slot="answer">
      <span>I'm the first answer!</span>
    </wp-fill>
    <wp-fill slot="answer">
      <span>I'm the second answer!</span>
    </wp-fill>
  </wp-slot>
  ```

  ```html
  <!-- Written by the developer -->
  <div wp-slot="answer">
    <!-- default content if no fills exist -->
    No answer
  </div>

  <!-- somewhere else in the document -->
  <span wp-fill="answer">I'm the first answer!</span>
  <span wp-fill="answer">I'm the second answer!</span>

  <!-- After processing -->
  <div wp-slot="answer">
    <span wp-fill="answer">I'm the first answer!</span>
    <span wp-fill="answer">I'm the second answer!</span>
  </div>
  ```

I think that's it for the atomic directives 🙂

@santosguillamot can you think of anything else?

> Although of course a bit more implicit. Something like
>
> ```html
> <div wp-do-something-based-on-criterion="context.some.information"></div>
> 
> <something-really-different maybe-with-some="attribute"
>   >abcd</something-really-different
> >
> ```

I don't think we will need that for the atomic directives, but it would be good to know if the system can support it easily or not.

---

By the way, I wonder if the correct behavior of `wp-show` should be to hide the node wrapper as well:

```html
<!-- Written by the developer -->
<div wp-show="myblock.falsyValue">Content...</div>

<!-- After processing -->
<template>
  <div wp-show="myblock.falsyValue">Content...</div>
</template>
```

What do you think?

adamziel:
> This means we need syntax like `getAttribute( 'wp-bind:*' )`.

How about `get_attributes()` that returns a `key => value` of all the tag attributes and allows you to handle any filtering you need?

Edit: @ockham already proposed it earlier and there's an [ongoing discussion](https://github.com/WordPress/block-hydration-experiments/pull/118#discussion_r1048863170)

wp-directives.php
adamziel: > you could end up ballooning the entire render process when a given tag has a lot of attributes

Tag Processor already does that. It's processing every tag in the document instead of short-circuiting that I'm worried about, but then again – if you want to find _all tags_ with a given directive, you'll end up processing _all tags_. 

An explicit list of directives still sound compelling – after all, you need to compare the attributes you'd get from `wp-data:*` against a set of supported names anyway.

Future optimization-wise, an early `strpos( $html, 'wp-data:' )` could indicate whether `$html` is even worth processing, whether the current tag is likely to be interesting, and how soon the processor can short-circuit when looking for an attribute.

wp-directives.php
dmsnell: In the absence of more confidence I'm way more comfortable with user-level shorthands than framework level shorthands.

> which will include all bookmarks that apply to it and which can then be used to modify it in a number of defined ways before stitching it back in?

Even this just screams for data sync issues. With all of the bookmarks and pointers contained within the `WP_HTML_Tag_Processor` everything stays in harmony. Granted, bookmarks can disappear and that's another form of the same problem, but at least then we're looking at one object which we have modified vs. birthing a new object, doing stuff to it, and then finding out later that the first object is now different and our updates invalid.

Here's one thing that's fine though: if you extract that inner content, do stuff to it, and then stitch it back in there's no problem and no cross-talk and no interference and no coupling.

That method hasn't been working for me with CSS selection because I need bookmarks back into the outer/parent region of the HTML, but for modifying content as I think you're doing here I don't think there are any issues.

wp-directives.php
dmsnell: > Tag Processor already does that. It's processing every tag in the document

It's not though since it only stores the offsets for where those attributes and their values start and end. If we approach a 2MB data-URI we will scan over it quickly and then record where it stops. Ballooning occurs if a single tag has hundreds of attributes, but even then the scaling is with the number of attributes, not with the length of their data.

It's only when calling `get_attribute()` that we allocate a new string and run that string through `html_entity_decode()`, which here would imply going _from_ doing that a few times in an input document _to_ doing it for every attribute on every tag.

I've been thinking about this and I think _at a minimum_ we'd want something to return attributes matching a prefix. That would allow us to ignore most things we don't care about and say instead, `get_attributes_by_prefix( 'wp-show:' )`, but even this makes me uncomfortable and makes me want to do things like say, you can only get these with a prefix N-characters long to avoid people calling `get_attributes_by_prefix( '' )`

If that's what someone wants to do I don't want to stop them, but I don't want to build a system that encourages the dangerous behavior, particularly because of how sensitive and important this one component in the system is. Nothing crazy, just want to make sure we actually need something we want and it's not simply a nice-to-have convenience that trades off predictability, guarantees, and performance.

SantosGuillamot:
> @SantosGuillamot can you think of anything else?

Not right now. I believe that covers the initial list we had in mind 🙂

wp-directives.php
adamziel: I'm hesitant about baking-in any attributes names filtering method as there will always be more a advanced use-case to cover. I'm now thinking about `get_attributes_names()` to let the consumer filter the already-allocated names without copying all the attributes values.

ockham: > I'm hesitant about baking-in any attributes names filtering method as there will always be more a advanced use-case to cover.

That makes sense; however, I'll point out that `get_attributes_by_prefix` aligns rather nicely with the HTML spec, since it can be used to e.g. get all dataset attributes (`data-`), or all attributes in a given namespace (e.g. `ng-`, `x-`, `wp-`) 😄 

Edit: The point being that `get_attributes_by_prefix` is arguably a bit less arbitrary as a filtering method.

adamziel: @ockham I do agree that `get_attributes_by_prefix()` seems like the best choice of a filtering method. My point is that I don't think we can avoid `get_attributes_names()` in the longer run and thus `get_attributes_by_prefix()` would be just a syntax sugar.

ockham: Here's an attempt to implement `get_attributes_by_prefix`: https://github.com/WordPress/gutenberg/pull/46672

ockham:
Update:

- I've added a `get_attributes_by_prefix` method (counterpart GB PR [here](https://github.com/WordPress/gutenberg/pull/46672); alternative `get_attribute names` [here](https://github.com/WordPress/gutenberg/pull/46685)).
- As the directive processor functions were getting more and more arguments, I decided to change their signature to accept a `WP_HTML_Tag_Processor` object reference instead (plus `$context`).
- The above two items then allowed me to implement (a very basic version of) `wp-bind` (https://github.com/WordPress/block-hydration-experiments/pull/118/commits/751027ca44d0505750e0d576fda2ef90d633173a).

wp-directives.php
dmsnell: > I don't think we can avoid get_attributes_names() in the longer run

why do you think this?

if that's the case though too, and we allow a zero-length prefix, that would accomplish the goal while still possibly encouraging people by default to only grab a subset.

ockham:
This branch is based on #100 and thus includes some fairly unrelated stuff (sample blocks etc). Furthermore, the files in its `src/html/` directory are in a bit of a chaotic state (combined from different open Gutenberg PRs).

I've thus decided to move development over to a fresh branch: #125, to make it easier to track and document any changes.

PR #119, opened 2022-12-14
WP Directives Plugin: Require Gutenberg plugin to avoid a fatal error

cbravobernal:
## Why

Right now, in the weird case that somebody installs the directives plugin without Gutenberg, they will get a fatal error, as we are using `WP_HTML_Tag_Processor`, which is declared in that plugin.
Screenshot 2022-12-14 at 16 33 04

## How
With GPT-3, of course
Screenshot 2022-12-14 at 16 39 57

## Testing

https://user-images.githubusercontent.com/37012961/207641442-f64d0d1f-5d53-4eb6-9b95-38e1468f43b9.mov

### DIFF
--- a/wp-directives.php
+++ b/wp-directives.php
@@ -10,8 +10,20 @@
  * Text Domain:       wp-directives
  */
 
+ // Check if Gutenberg plugin is active
+if (!function_exists('is_plugin_active')) {
+    include_once(ABSPATH . 'wp-admin/includes/plugin.php');
+}
+if ( ! is_plugin_active( 'gutenberg/gutenberg.php' ) ) {
+    // Show an error message
+    add_action( 'admin_notices', function() {
+		echo sprintf( '<div class="error"><p>%s</p></div>', __( 'This plugin requires the Gutenberg plugin to be installed and activated.', 'wp-directives' ) );
+    });
 
-
+    // Deactivate the plugin
+    deactivate_plugins( plugin_basename( __FILE__ ) );
+    return;
+}
 
 function wp_directives_loader()
 {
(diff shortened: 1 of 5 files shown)


SantosGuillamot (approved):
LGTM! A minor detail, it seems there are some different formatting in the functions.

Btw, not sure what is the status of the [Plugin Dependencies Feature Project](https://make.wordpress.org/core/2022/10/12/call-for-testing-plugin-dependencies/), but in the future we should use that I guess.

cbravobernal:
> Btw, not sure what is the status of the [Plugin Dependencies Feature Project](https://make.wordpress.org/core/2022/10/12/call-for-testing-plugin-dependencies/), but in the future we should use that I guess.

It seems that it is a plugin. You need to have it installed, so I guess having not won't prevent the fatal error.

>  it seems there are some different formatting in the functions.

I installed and applied WordPress PHP format so it's the same as standards. Also applied to the PHP files we had.

SantosGuillamot:
> It seems that it is a plugin. You need to have it installed, so I guess having not won't prevent the fatal error.

Yes, I think right now it is a plugin. But if I am not mistaken, they are planning to include in core. That's why I said that in the future, we could do this using that feature. It was just an informative comment, nothing to change 🙂 

> I installed and applied WordPress PHP format so it's the same as standards. Also applied to the PHP files we had.

Oh, I thought we already had that, sorry 😄 Thanks a lot for adding it!

SantosGuillamot (approved):

wp-directives.php
```diff
@@ -10,96 +10,110 @@
  * Text Domain:       wp-directives
  */
 
-function wp_directives_loader()
-{
+ // Check if Gutenberg plugin is active
+if ( ! function_exists( 'is_plugin_active' ) ) {
+	include_once( ABSPATH . 'wp-admin/includes/plugin.php' );
+}
+if ( ! is_plugin_active( 'gutenberg/gutenberg.php' ) ) {
```
luisherranz, line 17: @c4rl0sbr4v0, do you know what happens if Gutenberg is installed in a different folder?

cbravobernal: `is_plugin_active` uses the path to the plugin file relative to the plugins directory, I checked the code on Trac and is using a `get_option` function, so it should work. Gutenberg should be on `wp-contents/plugins`, otherwise it won't work, right? (I mean, a symlink should also work)

`ABSPATH` is a constant that returns the absolute path (sounds kind of obvious 😆 ) and while not recommended to access content folders, it should by fine when accessing Core folders.

One thing I don't love is having to include a core file to get `is_plugin_active` function, I guess I could find an alternative way to prevent this 🤔

luisherranz: Sorry, I meant if it's installed in a different folder within the plugins folder, like `wp-content/plugins/gunteberg-v15` for example.

cbravobernal: In that case it won't recognise Gutenberg as an installed and active plugin, and the notification will appear. The main page of the plugin stops working too [http://movies.local/wp-admin/gutenberg](http://movies.local/wp-admin/gutenberg)

I will take a look at it!

luisherranz: > I will take a look at it!

Nah, don't worry. It's not important and we probably won't need this. I was just curious 🙂

Issue #117, opened 2022-12-12
How to process `xpacket` in `toVdom`


SantosGuillamot:
We've found a couple of sites that have `xpacket` sections: [talsunovads.lv](https://href.li/?https://talsunovads.lv/) and [hlz.de](https://href.li/?http://hlz.de/).

Checking the [nodeType](https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType), it seems they are considered `nodeType` 7. As we are not handling that use case, it returns an error: `TypeError: Cannot read properties of undefined (reading 'length')`.

We have to decide what to do with these nodes. Should we do the same as with [CDATA nodes](https://github.com/WordPress/block-hydration-experiments/issues/109)? It seems a similar case, as they are included inside the SVGs.

DAreRodz:
> Should we do the same as with [CDATA nodes](https://github.com/WordPress/block-hydration-experiments/issues/109)? It seems a similar case, as they are included inside the SVGs.

It's a similar case indeed, but I think we should remove them in this case like we are doing with comments.

To be honest, I don't fully understand how browsers handle [Processing Instruction](https://developer.mozilla.org/en-US/docs/Web/API/ProcessingInstruction) nodes once they're instantiated. And the only thing I've found regarding `xpacket` instructions is that they work as opening and closing tags of [XMP](https://en.wikipedia.org/wiki/Extensible_Metadata_Platform) metadata sections in media files.

PR #121, opened 2022-12-16
Treat `nodeType` 7 like comments

SantosGuillamot:
As explained in #117 , there are some sites using `xpacket` inside SVG nodes, which returns a hydration error. As David suggests [here](https://github.com/WordPress/block-hydration-experiments/issues/117#issuecomment-1353029764), and following [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/ProcessingInstruction), I am filling this PR to treat them like comments and removing them.

I am planning to add tests for this and other use cases in other PR.


luisherranz (approved):
LGTM!

Issue #116, opened 2022-12-02
How to deal with `ref` attributes?


SantosGuillamot:
I was running the stress tests on some sites, and I found the same `hydrationError` in all of them: `TypeError: Cannot create property 'current' on string 'MobileMenu'` (the string varies between them).

I've been triaging it, and I believe this error occurs because they have a `ref` attribute in some HTML elements. As [`ref` is also used by Preact](https://preactjs.com/tutorial/05-refs/), I assume at some point it tries to do something like `ref.current === 'string'` and produces this error.

Some sites are adding `ref` because they use other libraries like Vuejs or plugins like [Curator.io](https://curator.io/), which also need the `ref` attribute.

This only happens when I run the stress testing script. If I paste the manual script into the browser console, I cannot reproduce it. I've checked the vDom with @DAreRodz, and the `ref` property is there, but it doesn't throw any error.

I tested 4000 sites and 8 of them had this issue:
- [kxbox.com](http://kxbox.com/): `TypeError: Cannot create property 'current' on string 'pc.kx.idx.banner1.ljxz'`
- [jhamiba.com](http://jhamiba.com/): `TypeError: Cannot create property 'current' on string 'MobileMenu'`
- [meidouya.com](http://meidouya.com/): `TypeError: Cannot create property 'current' on string 'MobileMenu'`
- [wpdaxue.com](http://wpdaxue.com/): `TypeError: Cannot create property 'current' on string 'MobileMenu'`
- [22vd.com](http://22vd.com/): `TypeError: Cannot create property 'current' on string 'nofollow'`
- [knewsmart.com](http://knewsmart.com/): `TypeError: Cannot create property 'current' on string 'MobileMenu'`
- [mfisp.com](http://mfisp.com/): `TypeError: Cannot create property 'current' on string 'MobileMenu'`
- [pro-q.it](http://pro-q.it/): `TypeError: Cannot create property 'current' on string 'image'`

---
Any idea about how should we approach this?

DAreRodz:
@SantosGuillamot, I've been taking a look at this issue, and I think we can ignore the `ref` attribute while generating virtual nodes, i.e., something like:

```diff
diff --git a/src/runtime/vdom.js b/src/runtime/vdom.js
index 5657128..b11cec9 100644
--- a/src/runtime/vdom.js
+++ b/src/runtime/vdom.js
@@ -24,6 +24,8 @@ export default function toVdom(node) {
[...]

PR #122, opened 2022-12-16
Ignore `ref` attr while generating virtual nodes

SantosGuillamot:
As @DAreRodz suggested [here](https://github.com/WordPress/block-hydration-experiments/issues/116#issuecomment-1352854882), I'm ignoring the `ref` attribute while generating virtual nodes. This seems to solve the hydration error that sites with `ref` attributes were returning.

I am planning to add tests for this and other use cases in other PR.


luisherranz (approved):
LGTM!

PR #123, opened 2022-12-16
Update stress testing with handling of `ref` and `xpacket`

SantosGuillamot:
After merging #121 and #122 , I'm applying the same changes in the stress testing branch. I basically:

- Merged the latest changes.
- Change the manual testing script.
- Prevent `xpacket` mutations from being registered in the DB.


luisherranz (approved):
👍

PR #103, opened 2022-11-16
🎨 Switch toVdom from array iteration to createTreeWalker

luisherranz:
I have been testing `createTreeWalker`, and it is consistently faster than the `for` we are using now: 50% in Chrome and 15% in Firefox and Safari.

This PR is not ready to merge yet. I have opened it up with my tests if anyone wants to do theirs as well. I have also recorded a video explaining what I have done so far:

<a href="https://www.loom.com/share/0b0966748b944c1cb858c45a10def67b">
    
  </a>

- https://www.loom.com/share/0b0966748b944c1cb858c45a10def67b
- https://excalidraw.com/#json=_2a96abSkBv_HWZsal2ri,XoSvfXho0gFrtutfak8Cjg
- https://stackblitz.com/edit/vitejs-vite-zyzjzm?file=index.html,main.js&terminal=dev
- https://jsbench.me/6flajfs1t9/1

I won't be able to continue with this PR in the next few days, so if anyone wants to, you can finish it: remove the tests, delete old code, etc. I haven't tested the algorithm with HTML comments either, so we would have to test that it works before merging it. We also have to start thinking about adding tests (unit and e2e), but maybe that can be done in another PR.

---

Thanks to @michalczaplinski for the HTML file generator 🙂

PR #124, opened 2022-12-20
Limit the hydration to block when client-side navigations are deactivated

### LINKED ISSUE #93: Limit the hydration of the Directives Hydration when the site doesn't support client-side navigations

luisherranz:
### What

This PR will limit the hydration so it only happens in the interactive blocks.

Fixes #93.

### Why

To avoid the cost of the Directives Hydration when it's strictly not necessary. Full explanation in #93.

### How

It keeps using the same `meta` tag mechanism to know if client-side navigations are enabled or not. If they are not, instead of doing the Directives Hydration, it looks for the interactive blocks and hydrates only those. Interactive blocks opt-in using a `supports.interactivity` property in their `block.json` and they are marked with a `wp-island` attribute in the HTML.

### Tasks

- [x] Add a block example
- [x] Add `wp-island` attribute to blocks with `supports.interactivity`
- [x] Limit the hydration to the blocks with the `wp-island` attribute
- [x] Add `wp-ignore` attribute
- [x] Add inner blocks to the blocks
- [x] Stop the hydration in the `wp-ignore` nodes
- [x] Add support for interactive block parents (as explained in #93)
- [x] Don't stop the hydration for interactive block parents
- [x] Fix syntax (switch from `prettier-php` to `phpcs`)
- [x] Don't hydrate interactive blocks that are inside interactive block parents twice
- [x] Support hydration of interactive blocks that are also inner blocks
- [x] Add e2e tests
- [x] Don't add `wp-island` and `wp-ignore` when client-side navigations are enabled
- ~~Don't stop in `wp-ignore` when client-side navigations are enabled~~

### Additional comments

- I've refactored the admin page because it was not saving the setting, I don't know why.
- I've fixed a problem in the `deepsignal` code, which was broken in the main branch.


SantosGuillamot:
Thanks for opening the Pull Request!

Something I was wondering about: Would it make sense, and would it be possible, to automatically detect if a block is using directives and add the `wp-interactive-block` attribute even if the block developer didn't add the `supports.interactivity` property in the `block.json`? Now that we are planning to use the WP_HTML_Tag_Processor, it seems feasible to me, but I am not sure if it would be safe or if it even makes sense. Any thoughts?

Btw, it could be something to do in the future and it may be out of the scope of this PR.

luisherranz:
> automatically detect if a block is using directives and add the wp-interactive-block attribute

Yes, it would be possible 🙂 

What is not possible is to differentiate between an "isolated" and a "parent" block. But maybe we could do the opposite and treat every interactive block as a parent block unless it specifies that it is isolated? That way it'd be safer to automatically detect the directives.

SantosGuillamot:
I've just watched again the video you shared in [the issue](https://github.com/WordPress/block-hydration-experiments/issues/93), and you mentioned it. I knew I had heard the idea somewhere 😄 

> But maybe we could do the opposite and treat every interactive block as a parent block unless it specifies that it is isolated?

Maybe not every interactive block but the ones including directives that could imply it needs inter-block communication like `wp-context`, `error-boundaries`, etc.? As a block developer, I believe I would prefer to use the `block.json` to opt-out and not having to include a property for each interactive block I create. It could also help for plugins modifying the markup of a specific block to add some directives and not have to modify the `block.json` as well.

luisherranz:
> the ones including directives that could imply it needs inter-block communication like wp-context, error-boundaries, etc.? 

Yeah, good point. That would require declaring it somehow on the directive declaration, but directive declarations will be used way less often so we can push that complexity there 🙂

> It could also help for plugins modifying the markup of a specific block to add some directives and not have to modify the block.json as well.

True!

---

Let's do that then, but in a subsequent PR.

luisherranz:
This should be ready for review.

<a href="https://www.loom.com/share/78003554808d492b819794170e95f38a">
    
  </a>

https://www.loom.com/share/78003554808d492b819794170e95f38a
https://excalidraw.com/#json=yPz4d84xWWLZR-KqTQ0_K,-FjTu31WIkjU4HSTDfh5jQ

Some notes:

- I didn't add tests for the PHP code yet.
- ~~We need to confirm that `querySelectorAll` returns parents first.~~ Confirmed by Mario here: https://github.com/WordPress/block-hydration-experiments/pull/124#issuecomment-1363744032
- We need to confirm that doing `dangerouslySetInnerHtml` doesn't recreate the HTML nodes.
- We should prevent `wpx` from being bundled inside each `view.js` file.
- I forgot to mention in the video that the [Playwright VS Code extension](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright) is awesome.
- Thanks to @c4rl0sbr4v0 for helping me find the bug in the admin page.
- Thanks to @darerodz for his code to add `wp-ignore` to the inner blocks.

luisherranz:
By the way, once this is merged, I want to add it to https://github.com/WordPress/block-hydration-experiments/pull/103 to compare the `querySelectorAll` with `createTreeWalker` to see if what Manu from Qwik said in [this tweet](https://twitter.com/manucorporat/status/1588422848678621184) is true in our use case as well.

> This is already done with createTreeWalker() which we found even faster than doing a querySelectAll()

SantosGuillamot:
I started taking a look at the PR. It looks great! I have a couple of comments so far:
* I was testing the blocks, as you do in the video, and it seems that when it is `isolated`, it only adds the `wp-ignore` attribute to the first children of the same block. Maybe we have to iterate somehow [here](https://github.com/WordPress/block-hydration-experiments/blob/limit-hydration/wp-directives.php#L166-L171)? For example, adding the following custom HTML in an isolated block doesn't seem to work as expected because only the paragraph gets the attribute:
```html
<p>Isolated block</p>
<div><wp-show when="state.dontShow">hide me</wp-show></div>
```

* Regarding the potential issue with `querySelectorAll` returning the children before the parent, [it seems the method returns the NodeList in document order](https://www.w3.org/TR/selectors-api/#queryselectorall): _"The querySelectorAll() methods on the [Document](https://www.w3.org/TR/selectors-api/#document), [DocumentFragment](https://www.w3.org/TR/selectors-api/#documentfragment), and [Element](https://www.w3.org/TR/selectors-api/#element) interfaces must return a [NodeList](https://www.w3.org/TR/selectors-api/#nodelist) containing all of the matching [Element](https://www.w3.org/TR/selectors-api/#element) nodes within the [subtrees](https://www.w3.org/TR/selectors-api/#subtrees) of the [context node](https://www.w3.org/TR/selectors-api/#context-node), in [document order](https://www.w3.org/TR/selectors-api/#document-order)."_

luisherranz:
> it only adds the wp-ignore attribute to the first children of the same block

Yes, the requirement for [blocks to have a single wrapper node](https://github.com/WordPress/block-hydration-experiments/commit/110dcbf5287dd46835fa2555460140b83f5dbe0d) extends to inner blocks as well. Maybe the Custom HTML block should have a default wrapper: `<div class="wp-block-custom-html">...</div>`.

But maybe something can be done with the [new capabilities of the `WP_HTML_Tag_Processor`](https://github.com/WordPress/gutenberg/pull/46018) to overcome this limitation.

> Regarding the potential issue with querySelectorAll returning the children before the parent, it seems the method returns the NodeList in document order

Wonderful. Thanks for checking it out, Mario!

wp-directives.php
```diff
@@ -101,22 +114,91 @@ function wp_directives_add_wp_link_attribute( $block_content ) {
...
 }
-add_action( 'wp_head', 'wp_directives_client_site_transitions_meta_tag', 10, 0 );
 
-/* User code */
+add_action(
+	'wp_head',
+	function () {
+		if ( wp_directives_get_client_side_transitions() ) {
+			echo '<meta itemprop="wp-client-side-transitions" content="active">';
+		}
```
cbravobernal: Can we move this to a named function? This way we keep consistency with all actions/filters.

SantosGuillamot (approved):
>Yes, the requirement for [blocks to have a single wrapper node](https://github.com/WordPress/block-hydration-experiments/commit/110dcbf5287dd46835fa2555460140b83f5dbe0d) extends to inner blocks as well. Maybe the Custom HTML block should have a default wrapper.

Okay, I didn't remember that part. Then everything looks great to me 🙂 Although it would be great if someone else can double check the code.

wp-directives.php
luisherranz: Sure. Done in https://github.com/WordPress/block-hydration-experiments/pull/124/commits/94f594fdb2dd25f047e138bfd5a6c3ff65825ea8 👍

wp-directives.php
```diff
@@ -101,22 +114,85 @@ function wp_directives_add_wp_link_attribute( $block_content ) {
...
 
-function wp_directives_client_site_transitions_meta_tag() {
-	if ( apply_filters( 'client_side_transitions', false ) ) {
+function wp_directives_get_client_side_transitions() {
+	static $client_side_transitions = null;
+	if ( is_null( $client_side_transitions ) ) {
+		$client_side_transitions = apply_filters( 'client_side_transitions', false );
+	}
+	return $client_side_transitions;
+}
```
cbravobernal, line 137: Can we here just call the function that gets the options instead of creating a new filter? `get_option()` by default uses filter and return false in case it does not exist.

Something like 
```php
$directive_plugins_settings = get_option('wp_directives_plugin_settings');
is_array($directive_plugins_settings) return $directive_plugins_settings['client_side_transitions'];
return $false;
```

luisherranz: It's very unlikely that we will start storing that option in a plugin's setting. It will likely be an experimental feature that will need to be turned on using the filter, so I'd rather keep it as it is 🙂

luisherranz:
Merging now so we can continue. If there's something broken, we'll fix it later 🙂

SantosGuillamot:
Btw, should we open a new issue to automate this with the HTML_Tag_Processor? Or is there a better place to keep track of that?

luisherranz:
Done! 

- https://github.com/WordPress/block-hydration-experiments/issues/126

PR #127, opened 2022-12-28
Fix wp-class and introduce the evaluate function

luisherranz:
### What

I improved the `wp-class` implementation because it was very basic. My idea was also to add unit tests, but instead I added more e2e tests. I also took the opportunity to improve the logic that resolves the `path`.

### Why

Because the implementation was so basic that it could fail with very simple cases.

### How

For `wp-class` I have introduced regular expressions. For the e2e tests I followed what I did in [my previous pull request](https://github.com/WordPress/block-hydration-experiments/pull/124). And for the logic that resolves `path`, I have introduced a new function `evaluate`.

I did a video to explain in detail each change:

<a href="https://www.loom.com/share/2fd4cbd8963f416da49b71c484d68b65">
    
  </a>

https://www.loom.com/share/2fd4cbd8963f416da49b71c484d68b65


DAreRodz (approved):
Looks good to me, @luisherranz! 👌

PS: I would have added some e2e tests for class names contained inside others (e.g., `"foo foo-bar"`) or just part of a class name (e.g. `foo` inside `"foo-bar foo-baz"`) to ensure that the regexp only handle whole class names. This is something you mentioned in the video, so  I was expecting these kinds of tests to be included. 🙂

luisherranz:
> I would have added some e2e tests for class names contained inside others (e.g., "foo foo-bar")

I'll add them before merging. Thanks, David!

michalczaplinski (approved):
Looks 💯 ! TIL about the Playwright VSCode extension!

luisherranz:
@darerodz: is https://github.com/WordPress/block-hydration-experiments/pull/127/commits/464b4eb6c4c7b38c4d3aad3400bac684733fa4ed enough? I don't fully understand what you mean by `foo inside "foo-bar foo-baz"`.

DAreRodz:
> @DAreRodz: is [464b4eb](https://github.com/WordPress/block-hydration-experiments/commit/464b4eb6c4c7b38c4d3aad3400bac684733fa4ed) enough? I don't fully understand what you mean by `foo inside "foo-bar foo-baz"`.

Thank @luisherranz! I think it's enough.

That test would have been just to ensure that the directive doesn't mess with other classes containing the one defined for the directive, even when that class is not there yet. But I guess that is already covered with the first test. 😊

luisherranz:
Merged then 🙂

PR #128, opened 2023-01-10
Switch to the deepsignal package

luisherranz:
## What

Just switch from the internal implementation of `deepSignal` to the [deepsignal](https://github.com/luisherranz/deepsignal/) package.

## Why

Because the package is feature complete, has versioning, tests, etc. We'll move it to the WordPress org in the future if WordPress ends up using it.

## How

Just remove the `deepSignal` code and adding `deepsignal` as a dependency.


michalczaplinski (approved):
Awesome 👍

PR #125, opened 2022-12-22
SSR: Experiment with `WP_HTML_Tag_Processor`

ockham:
Trimmed-down version of #118, based directly on the `main-wp-directives-plugin` branch.

- Adds PHP unit test infrastructure and CI.
- ~Adds `WP_HTML_Tag_Processor` files from Gutenberg `trunk` (i.e. everything from GB's `lib/experimental/html/` is copied to `src/html/`).~ _Now relies on the GB plugin to be installed and active, and uses its files directly._

We then add individual directives into `src/directives/`, and their corresponding unit tests into `phpunit/directives/`.

This should allow us to adopt a test-driven development approach, as we implement [all missing directives](https://github.com/WordPress/block-hydration-experiments/pull/118#issuecomment-1353407526), and to identify what other features we need in `WP_HTML_Tag_Processor` (also see #120).

✅  - Implemented
🟡 - Waiting for an upcoming Gutenberg release to include a missing feature
:x: - Blocked by missing feature in `WP_HTML_Tag_Processor`

| Name | Attribute directive | Tag directive | Missing feature |
|-------|-------------------|--------------|-----------------|
| `wp-context` | | ✅ | |
| `wp-show` | :x: | :x: | Wrapping content inside HTML tags with extra tags. E.g. `get_content_inside_balanced_tags()` from https://github.com/WordPress/gutenberg/pull/46345, and a matching `set_content_inside_balanced_tags()` (see https://github.com/WordPress/gutenberg/pull/47036). |
| `wp-bind` | ✅ | N/A | [`get_attribute_names_with_prefix()`](https://github.com/WordPress/gutenberg/pull/46840); landed in GB 15.0 |
| `wp-class` | ✅ | N/A | See `wp-bind`. |
| `wp-style` | ✅ | N/A | See `wp-bind`. Furthermore, see https://github.com/WordPress/gutenberg/issues/46887. (The latter isn't a blocker, though.) |
| `wp-text` | | | Replace content between HTML tags. See `wp-show`. |
| `wp-for` / `wp-each` | 
| `wp-html` | | | Replace content between HTML tags. See `wp-show`. |
| `wp-slot` & `wp-fill` |

### Testing Instructions

_Note that this PR currently requires at least GB v14.9. You might have to `npm run wp-env destroy` and restart it in order to update your GB version._

In addition, this PR currently uses `get_attribute_names_with_prefix` from https://github.com/WordPress/gutenberg/pull/46840, which hasn't been merged into GB yet.

[...]

PR #133, opened 2023-01-19
SSR: Add `wp-context`, `wp-bind`, `wp-class`, `wp-style`

ockham:
Implement Server-side rendering of those directives via `WP_HTML_Tag_Processor`.

Gutenberg 15.0 is the first version to include [`get_attribute_names_with_prefix()`](https://github.com/WordPress/gutenberg/pull/46840). This means that we now have everything we need to implement `wp-bind`, `wp-class`, and `wp-style`. This PR thus adds those directives, plus `wp-context`.

This also adds PHP unit test infrastructure and CI, in order to allow us to adopt a test-driven development approach for individual directives. Those "infrastructure" files are mostly copied directly from GB, with some minor modifications. Refer to individual commit messages to learn more.

When testing locally, note that you might have to update your `wp-env` environment (`npm run wp-env start -- --update`) to ensure that your environment is using the latest GB version.

To test locally, run `npm run test:unit:php`.

Based on the more experimental #125.


ockham:
@WordPress/frontend-dx I'd like to open this for review already -- it implements (basic versions of) a few directives and establishes some infrastructure.

src/directives/wp-bind.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+require_once __DIR__ . '/utils.php';
+
+function process_wp_bind( $tags, $context ) {
```
ockham, line 5: This is the interface I ended up with for the "directive processors".

- `$tags` is a `WP_HTML_Tag_Processor` object
- `$context` is a `WP_Directive_Context` ([see](https://github.com/WordPress/block-hydration-experiments/pull/133/files#diff-d5038f0e923a4b370eb61c523adc0baabd2c3187dfe7755b538b7e693e5e6649))

I previously tried to avoid passing a "whole" `WP_HTML_Tag_Processor`, but it turned out that we need pretty much everything that it offers, since we:

- Need to read its attribute values.
- Need to be able to change its attribute values (e.g. `wp-bind`, `wp-class`, `wp-style`).
- Need to be able to change the inner HTML of a given directive (e.g. `wp-text` and `wp-html`).

All that functionality is (or will be) provided by `WP_HTML_Tag_Processor` (or possibly a TBD `WP_HTML_Processor` that supports inner HTML modifications on top of tag-only modifications), so I'm pretty sure we need to pass the entire thing to the directive processors.

As for the second argument (`$context`): This might change, since we'll probably also want to be able to access other kind of contextual data (e.g. state), but I think it'll be easy enough to lump that together into one "context"/"state"/"data" arg later.

src/directives/utils.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+function get_from_context( $expr, $context ) {
```
ockham, line 3: This is currently only reading object paths, based on the current `$context`.

E.g. if `$expr` is `context.myblock.isOpen`, and `$context` is `array( 'myblock' => array( 'isOpen' => true ) )`, then `get_from_context( $expr, $context )` will indeed return `true`.

In the long run, we might want to support more complex expressions, e.g. `context.myblock.isOpen && ! context.otherBlock.isOpen`. This means we'll have to implement an expression parser.

src/directives/wp-bind.php
```diff
@@ -0,0 +1,31 @@
...
+		return;
+	}
+
+	/**
+	 * A `wp-bind` *tag* doesn't really make sense.
+	 * What would be the point of e.g. `<wp-bind:src="image.png">?
+	 */
+	if ( 'WP-BIND' === $tags->get_tag() ) {
+		return;
+	}
```
luisherranz, line 16: I'm not sure if it'd be better to divide the definition of directive attributes and directive tags into different functions as we do in JavaScript. Something like `process_wp_bind_tag` and `process_wp_bind_attribute`.

What do think, @ockham?

wp-directives.php
```diff
@@ -196,3 +204,36 @@ function bhe_inner_blocks( $parsed_block, $source_block, $parent_block ) {
...
+
+	$context = new WP_Directive_Context;
+	while ( $tags->next_tag( array( 'tag_closers' => 'visit' ) ) ) {
+		$tag_name = strtolower( $tags->get_tag() );
+		if ( array_key_exists( $tag_name, $directives ) ) {
+			call_user_func( $directives[ $tag_name ], $tags, $context );
+		} else {
+			// Components can't have directives (unless we change our mind about this).
+			foreach ( $directives as $directive => $directive_processor ) {
+				$attributes = $tags->get_attribute_names_with_prefix( $directive );
```
luisherranz, line 227: I wonder what would be more performant, doing `get_attribute_names_with_prefix` once per directive attribute, or just doing it once to get all the attributes and search for the directives ourselves.

Have you thought about it, @ockham?

luisherranz (approved):
This looks great! Thanks @ockham 🙂👏

I've left a couple of comments below.

src/directives/wp-bind.php
ockham: Yeah, I think that might make sense 🤔 Some directives even only make sense as _either_ attribute (e.g. `wp-bind`, `wp-class`, and `wp-style`) _or_ tag, but not as both. Keeping track of the context stack separately for `wp-context` tags and attributes also makes sense.

Originally, I liked the simplicity of having a map of directive names and corresponding `process_wp_*` functions, but that's not really a strong counterargument, so I think I'll go with separate processor functions for tags and attributes 👍 

(I'd like to do that in a follow-up to this PR if that's okay 😊 )

wp-directives.php
ockham: I hadn't before, but I like the idea! Happy to work on that (in another follow-up 😊)

src/directives/wp-bind.php
luisherranz: > (I'd like to do that in a follow-up to this PR if that's okay 😊 

Oh, don't worry too much about it, we can define the API to create directives later, and we will try to match the JS and PHP functions. Not important. I just wanted to know about your reasoning.

> Keeping track of the context stack separately for wp-context tags and attributes also makes sense.

What do you mean by this? 

_sorry for my late reply!_

wp-directives.php
luisherranz: I don't want to be a doomsayer, but wouldn't doing that defeat the purpose of introducing the `with_prefix` in the first place? 😆

luisherranz: > Happy to work on that (in another follow-up 😊)

By the way, I don't know the internals of `get_attribute_names_with_prefix`. I was just wondering which method would be more performant. I don't know if getting all the attributes at once is.

PR #135, opened 2023-01-20
Router: Use `window.location` to clean the URL

DAreRodz:
## What

Fixes a bug in `navigate` that made relative links fail.

## Why

We were using an arbitrary URL to clean URLs.

https://github.com/WordPress/block-hydration-experiments/blob/0c2d3d9bd7673c466ada815d72bdc53889f5838e/src/runtime/router.js#L15

As that arbitrary URL didn't contain any path, the relative links of WP sites that live in a particular path (e.g., `href="all-products"` in `https://domain.example/shop/`) stopped working because the site path (`/shop/`) was not included in the final URL.

## How

Using `window.location` to calculate the URL to navigate instead of an arbitrary URL.

### DIFF
--- a/src/runtime/router.js
+++ b/src/runtime/router.js
@@ -12,7 +12,7 @@ const stylesheets = new Map();
 // Helper to remove domain and hash from the URL. We are only interesting in
 // caching the path and the query.
 const cleanUrl = (url) => {
-	const u = new URL(url, 'http://a.bc');
+	const u = new URL(url, window.location);
 	return u.pathname + u.search;
 };
 


SantosGuillamot (approved):
It looks fine to me. To give a bit more context, this was detected [here](https://github.com/woocommerce/woocommerce-blocks/pull/8220#issuecomment-1397238113):

### Before this PR

If you have a Query Loop with Client Side navigations inside a page, the numbers in the pagination are doing navigating to the homepage, while the "Next Page" button is working fine. The only difference between them are the relative links:
* Numbers: Relative link points to `?query-0-page-2`.
* Next Page button: Relative link points to `/page-slug/?query-0-page-2`.

In the main branch right now, this causes that **numbers link is prefetching `https:://mysite.com?query-0-page-2` and doing a client-side navigation to the homepage**, which is not correct.

### After this PR

With this PR, the relative links remain the same, but **both are prefetching and doing client side navigation to the proper page**.

DAreRodz:
Thanks for the review, @SantosGuillamot. 😊

PR #139, opened 2023-01-26
Make `wp-context` inheritable and extensible

### LINKED ISSUE #137: Merge inherited `wp-context` properties from a parent instance
If a `wp-context` directive is nested under another `wp-context` instance, the properties defined in the parent context should also be available inside the nested context.

```jsx
const WpContextExample = () => {
  const contextA = { a: 'foo' };
  const contextB = { b: 'bar' };

  return (
    <div wp-context={JSON.stringify(contextA)}>
      <div wp-context={JSON.stringify(contextB)}>
        {/* `context.a` and `context.b` should be accessible here */}
      </div>
    <div>
  );
}
``` 

Right now, only `context.b` is accessible in that scope because `contextB` replaces `contextA` entirely.

We need to implement `wp-context` so that all the previous properties that don't collide are preserved, even when they are nested.

In addition, we need to agree on how [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array) objects would be merged. We can either append the elements or replace them according to their index.

DAreRodz:
## What

(Re) implement the `wp-context` directive so nested contexts can inherit properties from the parent context. Inherited properties can also be shadowed if redeclared in the child context.

## Why

At this moment, a child context replaces the parent context entirely, but contexts should be inheritable and extensible to cover most use cases. 

## How

Implementing a `mergeDeepSignals` function that copies signals from the parent context only if they're not already defined in the child context. That would allow, for example, changes in inherited properties from the child context to be reflected in the parent one, except in the case those properties are shadowed.

### DIFF
--- a/e2e/directives-context.html
+++ b/e2e/directives-context.html
@@ -57,6 +57,14 @@
                     <!-- rendered during hydration -->
                 </pre>
 				<button
+					data-testid="child prop1"
+					name="prop1"
+					value="modifiedFromChild"
+					wp-on:click="actions.updateContext"
+				>
+					prop1
+				</button>
+				<button
 					data-testid="child prop2"
 					name="prop2"
 					value="modifiedFromChild"
@@ -71,6 +79,14 @@
 					wp-on:click="actions.updateContext"
 				>
 					prop3
+				</button>
+				<button
+					data-testid="child obj.prop4"
+					name="obj.prop4"
+					value="modifiedFromChild"
+					wp-on:click="actions.updateContext"
+				>
+					obj.prop4
 				</button>
 				<button
 					data-testid="child obj.prop5"
--- a/e2e/directives-context.html
+++ b/e2e/directives-context.html
@@ -1,0 +1,98 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<title>Directives -- wp-context</title>
+		<meta itemprop="wp-client-side-transitions" content="active" />
+	</head>
+
+	<body>
+		<div
+			wp-context='{ "prop1":"parent","prop2":"parent","obj":{"prop4":"parent","prop5":"parent"},"array":[1,2,3] }'
+		>
+			<pre
+				data-testid="parent context"
+				wp-bind:children="derived.renderContext"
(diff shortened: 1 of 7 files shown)


src/runtime/wpx.js
```diff
@@ -16,6 +16,16 @@ export const deepMerge = (target, source) => {
...
+export const mergeDeepSignals = (target, source) => {
+	for (const k in source) {
+		if (typeof peek(target, k) === 'undefined') {
+			target[`$${k}`] = source[`$${k}`];
+		} else if (isObject(peek(target, k)) && isObject(peek(source, k))) {
+			mergeDeepSignals(target[`$${k}`].peek(), source[`$${k}`].peek());
+		}
+	}
+};
+
```
luisherranz: Why is the reasoning to put this here instead of in the same file as the `wp-context` directive definition (which will eventually be moved to its own file)?

e2e/directives-context.html
```diff
@@ -0,0 +1,114 @@
...
+				wp-bind:children="derived.renderContext"
+			>
+                <!-- rendered during hydration -->
+            </pre>
+			<button
+				data-testid="parent prop1"
+				name="prop1"
+				value="modifiedFromParent"
+				wp-on:click="actions.updateContext"
+			>
```
luisherranz, line 23: Using attributes as function arguments is smart. You are always coming up with interesting patterns, David :smile:

luisherranz (approved):
Beautiful. I think it's ready to merge 🙂

Oh, and thanks for starting to use TypeScript, by the way.

src/runtime/wpx.js
DAreRodz: I thought it could be useful for other (custom) directives... Yeah, well, the actual reason was that I just saw `deepMerge` defined and exported there, and I supposed I should do the same. 😅 

No worries, I can move it to `directives.js`.

luisherranz: Yeah, I think it's small and specific enough to belong to `wp-context`. Thank you! 🙂

PR #140, opened 2023-01-26
SSR: Handle attribute directives separately from tag directives

ockham:
Follow-up to #133, addressing the following feedback:

> I'm not sure if it'd be better to divide the definition of directive attributes and directive tags into different functions as we do in JavaScript. Something like `process_wp_bind_tag` and `process_wp_bind_attribute`.

(https://github.com/WordPress/block-hydration-experiments/pull/133#discussion_r1084153104)

and

> I wonder what would be more performant, doing `get_attribute_names_with_prefix` once per directive attribute, or just doing it once to get all the attributes and search for the directives ourselves.

(https://github.com/WordPress/block-hydration-experiments/pull/133#discussion_r1086352740)

- Create new subfolders `attributes` and `tags` inside both `src/directives/` and `phpunit/directives/`.
- In `wp-directives.php`, use separate arrays for attribute and tag directives.
- Remove now-obsolete code and tests from attribute directive processors (that would've dealt with being called on a tag directive).
- Differentiate `process_wp_context_tag` and `process_wp_context_attribute`.
- Disable unit test for `wp-context` attribute directive, as it's not implemented yet.


DAreRodz (approved):
Everything OK codewise. Approved! ✅ 

By the way, I see that you started calling directives to both tags and attributes. The idea I had in that regard was that just the attributes were _directives_, while the tags were simply _components_ (like in _Web Components_).

I mean, the names you used are OK. I guess we'll reach a consensus with the terminology anytime soon, haha. 😄

ockham:
> Everything OK codewise. Approved! ✅

Thank you very much!
 
> By the way, I see that you started calling directives to both tags and attributes. The idea I had in that regard was that just the attributes were _directives_, while the tags were simply _components_ (like in _Web Components_).

Yeah, this is something that I believe I've discussed with @luisherranz (I thought it was at #118 but I can't find it right now) -- it seemed to me like Luis preferred that nomenclature over directive/component. I'm fine with either -- maybe "component" is a bit more likely to be confusing, since it's used in so many different contexts 🤔 

> I mean, the names you used are OK. I guess we'll reach a consensus with the terminology anytime soon, haha. 😄

Indeed! 😄

luisherranz:
It was actually @dmsnell the first one who started using it. I adopted it right away because I thought it makes much more sense. I think others (@santosguillamot? @michalczaplinski?) have also been using it lately.

> I guess we'll reach a consensus with the terminology anytime soon, haha.

You're totally right. We haven't discussed it anywhere. Sorry about that, David.

I'll add it to the Tracking Issue:

- https://github.com/WordPress/block-hydration-experiments/issues/80

PR #141, opened 2023-01-26
SSR: Add `wp-show` attribute directive processor

ockham:
Implement the `data-wp-show` directive, which is supposed to transform code like

```html
<div data-wp-show="falsyValue" class="my-class">
  <p>Children</p>
</div>
```

into

```html
<template data-wp-show="falsyValue">
  <div class="my-class">
    <p>Children</p>
  </div>
</template>
```

per [this discussion](https://github.com/WordPress/block-interactivity-experiments/issues/159#issuecomment-1458572890). (For truthy values of `data-wp-show`, it does not modify the HTML at all.)

Note that the behavior to move the `data-wp-show` attribute from the wrapped tag to the wrapping one was implemented in https://github.com/WordPress/block-interactivity-experiments/pull/141/commits/da195d73f469e7d9b39e228513eeeebab99d5b5e. If we decide that we'd rather keep the attribute on the wrapped tag, we can simply revert that commit.


phpunit/directives/tags/wp-show.php
```diff
@@ -0,0 +1,64 @@
...
+		$markup = <<<EOF
+			<wp-show when="context.myblock.open">
+				I should be shown!
+			</wp-show>
+EOF;
+		$tags = new WP_Directive_Processor( $markup );
+		$tags->next_tag();
+
+		$context_before = new WP_Directive_Context( array( 'myblock' => array( 'open' => true ) ) );
+		$context = $context_before;
```
gziolo: Should it get cloned here? Otherwise, it’s the same object referenced.
```suggestion
		$context = clone $context_before;
```

or maybe the initial context should get assigned to the variable and used in the assertion later.

gziolo (commented):
All those directives are neat!

phpunit/directives/tags/wp-show.php
ockham: Ah, good spot! 😅 I have a few more of those in other unit tests, will file a PR to update them.

ockham:
Per https://github.com/WordPress/block-hydration-experiments/issues/152#issuecomment-1448482028, I'll update this PR to turn the `wp-show` tag directive into an attribute directive.

ockham:
I'm considering removing the `WP_Directive_Processor` I've added here, and rebasing this PR on #158 instead to use that PR's `WP_Directive_Processor` (yep, same name). This is currently blocked ([see](https://github.com/WordPress/block-hydration-experiments/pull/158#issuecomment-1452437543)) but should stop being a problem next week.

We could then rewrite the directive like so:

```diff
diff --git a/src/directives/attributes/wp-show.php b/src/directives/attributes/wp-show.php
index 89ed612..82b28cc 100644
--- a/src/directives/attributes/wp-show.php
+++ b/src/directives/attributes/wp-show.php
@@ -3,25 +3,19 @@
 require_once __DIR__ . '/../utils.php';
 
 function process_wp_show( $tags, $context ) {
-	if ( ! $tags->is_tag_closer() ) { // TODO: Exclude void and self-closing!
-		set_bookmark_for_directive( $tags, 'wp-show' );
+	if ( $tags->is_tag_closer() ) {
 		return;
 	}
 
-	$end = 'wp-show-closer';
-	$tags->set_bookmark( 'wp-show-closer' );
-	$start = seek_bookmark_for_directive( $tags, 'wp-show' );
-
 	$value = $tags->get_attribute( 'wp-show' );
-	if ( null !== $value ) {
-		$show = evaluate( $value, $context->get_context() );
+	if ( null === $value ) {
+		return;
+	}
 
-		if ( ! $show ) {
-			$content = $tags->get_content_inside_bookmarks( $start, $end );
-			$tags->set_content_inside_bookmarks( $start, $end, '<template>' . $content . '</template>' );
-		}
+	$show = evaluate( $value, $context->get_context() );
+	if ( ! $show ) {
+		$content = $tags->get_inner_html();
+		echo $content;
+		$tags->set_inner_html( '<template>' . $content . '</template>' );
 	}
-	$tags->seek( $end );
-	$tags->release_bookmark( $start );
-	$tags->release_bookmark( $end );
 }

```

dmsnell:
is there any foreseeable need for `set_outer_html()`? from what @adamziel discovered it seems like that one creates problems with the tag processor, but `set_inner_html()` doesn't. if we don't have to solve that problem I'd be in favor of letting it sit unsolved.

luisherranz:
Not having `set_outter_html` would limit our capability of wrapping the whole tag with `<template>` in `wp-show`:

From this:

```html
<div data-wp-show="false" class="my-class">
  <p>Children</p>
</div>
```

Doing this:

```php
$p->set_outter_html( '<template>' . $p->get_outter_html() . '</template>' );
```

```html
<template>
  <div data-wp-show="false" class="my-class">
    <p>Children</p>
  </div>
</template>
```

Instead of this:

```php
$p->set_inner_html( '<template>' . $p->get_inner_html() . '</template>' );
```

```html
<div data-wp-show="false" class="my-class">
  <template>
    <p>Children</p>
  </template>
</div>
```

We haven't decided about it yet, but yesterday [I shared an example](https://github.com/WordPress/block-hydration-experiments/issues/159#issuecomment-1452377597) of something that (I think) would not be possible with the `set_inner_html` option.

- https://github.com/WordPress/block-hydration-experiments/issues/159

This case could work with a "wrap tag" API as well, if that would make things easier.

---

It would also leave some leftovers when using `wp-fill`.

For example, this code:

```html
<table wp-slot="data"></table>

<template wp-fill="data">
  <tr>
    <td wp-text="id"></td>
    <td wp-text="firstName"></td>
    <td wp-text="lastName"></td>
  </tr>
</template>
```

Would end up as this using `$p->set_inner_html('')` to remove the fill once it has been moved to the correct place:

```html
<table wp-slot="data">
  <tr>
    <td wp-text="id"></td>
    <td wp-text="firstName"></td>
    <td wp-text="lastName"></td>
  </tr>
</table>

<template wp-fill="data"></template>
```

Leaving the empty template hanging around, but it doesn't affect the page.

This case could work with a "remove tag" API as well, if that would make things easier.

---

The rest of the core directives we have in mind right now don't seem to need anything similar to `set_outter_html`.

So at this moment, it depends on what we decide for `wp-show`. We need to keep investigating.

SantosGuillamot:
> This case could work with a "wrap tag" API as well, if that would make things easier.

Not sure if you meant this, but I suggested a [new approach](https://github.com/WordPress/block-hydration-experiments/issues/159#issuecomment-1453509805) to handle `wp-show` that wouldn't require `set_outer_html` if I am not mistaken: Restrict the `wp-show` directive to be used inside `<template>`. The same way you are using `wp-fill` in your examples. As mentioned there, I am not convinced about it, but I think you'll know better than me if it makes sense or not.

luisherranz:
It makes sense indeed. I've answered you on the other issue:

- https://github.com/WordPress/block-hydration-experiments/issues/159#issuecomment-1458572890

ockham:
Rebased on #169.

ockham:
We've [decided](https://github.com/WordPress/block-interactivity-experiments/issues/159#issuecomment-1458572890) that we'd like `wp-show` to work as follows:

> When `wp-show` is false:
> - Wrap it with `<template>` in the server (maybe also move the `wp-show` to the template).
> - Remove it from the DOM in the client.
>
> In the client, having the DOM elements wrapped by `<template>` is not useful. I mean, we already have the static virtual nodes that represent that HTML in memory, we don't need them in the DOM anymore.

I.e. we'd like the server-side rendered `wp-show` directive to return something like this:

```html
<template>
  <div data-wp-show="false" class="my-class">
    <p>Children</p>
  </div>
</template>
```

(with `data-wp-show="false"` possibly moved to `<template>`)

Our directive processor (#169) currently doesn't support `set_outer_html`, and it's been found that the latter could be somewhat tricky to implement.

As a consequence, I'd like to try implementing a `wrap_in_tag` (or so) method first, which should hopefully be easier, since it's not as potentially "destructive" to the current tag as `set_outer_html`.

ockham:
> I'd like to try implementing a `wrap_in_tag` (or so) method first, which should hopefully be easier, since it's not as potentially "destructive" to the current tag as `set_outer_html`.

Done with @michalczaplinski in https://github.com/WordPress/block-interactivity-experiments/pull/141/commits/39b2e1b0ab25982475d0eb47e822377ac7be2762 (but needs more polish 🇵🇱). Updated `wp-show` to use it in https://github.com/WordPress/block-interactivity-experiments/pull/141/commits/d3b36fa30a2066a67d7e097a5f33bd328b669877.

src/directives/class-wp-directive-processor.php
```diff
@@ -123,6 +123,75 @@ public function get_balanced_tag_bookmarks() {
...
+	 * element in the argument tag.
+	 *
+	 * Note that the internal pointer will continue to point to the same tag as before
+	 * calling the function.
+	 *
+	 * @param string $tag An HTML tag, specified in uppercase (e.g. "DIV").
+	 * @return string|false The name of a bookmark pointing to the wrapping tag opener
+	 *                      if successful; false otherwise.
+	 *
+	 * @todo Allow passing in tags with attributes, e.g. <template id="abc">?
```
ockham, line 148: Note that we're only accepting an uppercase HTML tag such as `DIV` or `TEMPLATE` as the `$tag` argument. We don't accept it wrapped in angle brackets, or with attributes (`<template attr="value>`), as that would require us to apply HTML parsing to the `$tag` argument itself -- and would allow for many more edge cases for which we'd need to decide how to behave (E.g. what if we pass two tags (`<template><div>`) -- do we wrap in both? What if one of them is void (`<template>`)? What if there's a basic syntax error (`<template`)?)

We can eschew most of these questions by only allowing a tag name; handling of attributes can be done after the fact and separately in `process_wp_show`.

SantosGuillamot:
I still have to review it properly, and I am not sure if I am doing something wrong, but I'm getting a weird HTML structure when the value is false.

For this HTML:
```html
<div data-wp-show="falseValue">
    <span>WP Show</span>
</div>
```

I'm getting this from the server:
```html
<template>
  <template data-wp-show="falseValue">
    <template>
      <div>
        <span>WP Show</span>
      </div>
    </template>
  </template>
</template>
```

Is it only me? As I said, I may be doing something wrong as I just quickly tested it.

ockham:
Hmm, weird 🤔 It's definitely possible that the directive isn't quite working correctly (as this one requires some things that we didn't previously need in other directives, such as moving the internal pointer around).

However, I couldn't reproduce when modifying the unit test:

```diff
diff --git a/phpunit/directives/attributes/wp-show.php b/phpunit/directives/attributes/wp-show.php
index f9314c1..3f812f5 100644
--- a/phpunit/directives/attributes/wp-show.php
+++ b/phpunit/directives/attributes/wp-show.php
@@ -34,7 +34,7 @@ class Tests_Directives_WpShow extends WP_UnitTestCase {
        }
 
        public function test_directive_wraps_content_in_template_if_when_is_false() {
-               $markup = '<div data-wp-show="context.myblock.open">I should not be shown!</div>';
+               $markup = '<div data-wp-show="context.myblock.open"><span>I should not be shown!</span></div>';
 
                $tags = new WP_Directive_Processor( $markup );
                $tags->next_tag();
@@ -46,7 +46,7 @@ class Tests_Directives_WpShow extends WP_UnitTestCase {
                $tags->next_tag( array( 'tag_closers' => 'visit' ) );
                process_wp_show( $tags, $context );
 
-               $updated_markup = '<template data-wp-show="context.myblock.open"><div >I should not be shown!</div></template>';
+               $updated_markup = '<template data-wp-show="context.myblock.open"><div ><span>I should not be shown!</span></div></template>';
 
                $this->assertSame( $updated_markup, $tags->get_updated_html() );
                $this->assertSame( $context_before->get_context(), $context->get_context(), 'data-wp-show directive changed context' );
```

I also tried adding more 

```
		$tags->next_tag( array( 'tag_closers' => 'visit' ) );
		process_wp_show( $tags, $context );
```

pairs to emulate iterating through the entire HTML, but that also didn't cause the test to fail 🤔 

I'll have another look next week! Hopefully we can find a minimal test case that we can turn into a unit test to guard against this kind of error 🤞

SantosGuillamot:
I still have to review the code properly, but I'm able to reproduce the issue with the following code:

```php
function add_wp_show($content)
{
	return $content . "<div data-wp-context='{ \"isOpen\": false }'><div data-wp-show='context.isOpen'><span>WP Show</span></div></div>";
}
add_filter(
	'render_block_core/post-title',
	'add_wp_show',
	10,
	1
);
```

ockham:
Thank you Mario!

My hunch is that it’s happening because the directive SSR happens at block level, which means it can run multiple times for nested blocks, as David [pointed out a while ago](https://href.li/?https://github.com/WordPress/block-interactivity-experiments/issues/148).

I'll try to add a basic unit test to cover this and see if it fails 😊

ockham:
> I'll try to add a basic unit test to cover this and see if it fails 😊

It's a bit tricky to distill this into a minimal test case -- still working on it.

@SantosGuillamot In the meantime, could you try if this patch (applied on top of this PR's branch) fixes the issue?

```diff
diff --git a/wp-directives.php b/wp-directives.php
index e4be9f3..5187b2a 100644
--- a/wp-directives.php
+++ b/wp-directives.php
@@ -230,7 +230,7 @@ function process_directives_in_block( $block_content ) {
        return $tags->get_updated_html();
 }
 add_filter(
-       'render_block',
+       'the_content',
        'process_directives_in_block',
        10,
        1
```

SantosGuillamot:
> @SantosGuillamot In the meantime, could you try if this patch (applied on top of this PR's branch) fixes the issue?

I quickly tested it, and it doesn't seem to work, but I believe `the_content` just returns the Post content, and I'm using most of the blocks outside of the Post content (in a template). I can try tomorrow to include those blocks in the post content itself.

ockham:
> I quickly tested it, and it doesn't seem to work, but I believe `the_content` just returns the Post content, and I'm using most of the blocks outside of the Post content (in a template). I can try tomorrow to include those blocks in the post content itself.

Thank you! Yeah, that makes a lot of sense.

I'll need to keep looking for a better filter to use (that includes templates and content); or alternatively to continue doing it on a per-block basis (which might also have some advantages), in which case I'd need to find a way to avoid processing directives twice for nested blocks 🤔

dmsnell:
@ockham this is something we should talk about this upcoming week, but I wonder how bad it would be to simply add a new attribute to prevent double-processing.

```php
$p->set_attribute( 'data-wp-has-processed', true );
```

Then we can bail if we've already set that.

```php
if ( $p->get_attribute( 'data-wp-has-processed' ) ) {
    continue;
}
```

in any case this might be a reasonable method to continue exploring until a clearly better solution is found, and this would mean we don't have to rely on people _doing it right_ by calling this in the appropriate places and not in others. I suspect there's an open list of places someone might call the processor multiple times that we don't anticipate.

luisherranz:
> someone might call the processor multiple times

I don't think devs should be manually calling the directives processor at all. It should be called only by Gutenberg.

dmsnell:
> I don't think devs should be manually calling the directives processor at all. It should be called only by Gutenberg.

For the most part I agree, but my comment is more about finding a safe solution anticipating that people will. If we can build that safety into the system such that it doesn't collapse when people do something we didn't anticipate or intent, I think that's responsible on our part; so this was only to suggest that a kind of latch mechanism solves our own needs plus it covers the other likely cases that we don't normatively want, vs. a kind of internal cache that would serve our purpose but be lost if someone from the outside interacted with it.

ockham:
I think that #215 should finally unblock this PR 🤞

phpunit/directives/wp-directive-processor.php
```diff
@@ -125,4 +125,55 @@ public function test_set_inner_html_invalidates_bookmarks_that_point_to_replaced
...
+
+	public function test_wrap_in_tag_wraps_balanced_tag_correctly_when_called_after_add_class() {
+		$tags = new WP_Directive_Processor( self::HTML );
+
+		$tags->next_tag();
+		$tags->add_class( 'foo' );
+		$tags->next_tag( 'section' );
+		$tags->wrap_in_tag( 'TEMPLATE' );
+		$this->assertSame( '<div class="foo">outside</div><template><section><div><img>inside</div></section></template>', $tags->get_updated_html() );
+		$this->assertSame( 'SECTION', $tags->get_tag() );
```
ockham: Note that this is currently [failing](https://github.com/WordPress/block-interactivity-experiments/actions/runs/4733465266/jobs/8401035058?pr=141) (but fixed by #215). The error is:

```
2) WP_Directive_Processor_Test::test_wrap_in_tag_wraps_balanced_tag_correctly_when_called_after_add_class
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'SECTION'
+'TEMPLATE'

/var/www/html/wp-content/plugins/block-interactivity-experiments/phpunit/directives/wp-directive-processor.php:147
```

Funny enough, if we change the `foo` to `fooooooo`, the error is different:

```
2) WP_Directive_Processor_Test::test_wrap_in_tag_wraps_balanced_tag_correctly_when_called_after_add_class
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'SECTION'
+'DIV'

/var/www/html/wp-content/plugins/block-interactivity-experiments/phpunit/directives/wp-directive-processor.php:147
```

(I.e. we're now at the `<div>` rather than the `<section>`.)

src/directives/class-wp-directive-processor.php
```diff
@@ -131,6 +131,84 @@ public function get_balanced_tag_bookmarks() {
...
+		if ( self::is_html_void_element( $tag ) ) {
+			// _doing_it_wrong(
+			// __METHOD__,
+			// __( 'Cannot wrap HTML in void tag.' ),
+			// '6.3.0'
+			// );
+			return false;
+		}
+
+		$this->get_updated_html(); // Apply potential previous updates.
```
dmsnell, line 164: After applying the fix in ~dmsnell/wordpress-develop#8~ WordPress/wordpress-develop#4345 and removing this call to `get_updated_html()` my test case, where I reproduced at least some of the errors we found in this PR, succeeded.

dmsnell: And this is the test I used to reproduce…

```php
public function blahblahblah() {
	$p = new WP_Directive_Processor( '<div><img hidden></div>' );

	$p->next_tag();
	$p->next_tag('img');
	$p->set_attribute( 'inert', true );
	$p->remove_attribute( 'hidden' );
	$b = $p->wrap_in_tag( 'figure' );
	$this->assertSame( 'IMG', $p->get_tag() );
	$p->seek( $b );
	$this->assertSame( 'FIGURE', $p->get_tag() );
	$p->next_tag();
	$this->assertSame( 'IMG', $p->get_tag() );

	$this->assertSame( '<div><figure><img inert ></figure></div>', $p->get_updated_html() );
}

```

ockham: Thank you!

I just tested your [patch](https://github.com/WordPress/wordpress-develop/pull/4345) (plus the removal of that `get_updated_html()` call inside `wrap_in_tag()`, and my changes from #215) against the Movies Demo. Unfortunately, the [issue I found there](https://github.com/WordPress/block-interactivity-experiments/pull/215#issuecomment-1513162325) still persists 😕 

On a sidenote -- it strikes me as odd that we need to remove a `get_updated_html()` call. Since I've conceptually thought of it as a flushing method, I was silently assuming that there was no harm in calling it. Am I mistaken here? Or is there still something wrong with it that we need to fix?

ockham: Also, the tests that are currently failing on this branch are still failing with your patch applied and that `get_updated_html` call removed 😕 

To verify that your patch was properly picked up, I copied the test from https://github.com/WordPress/wordpress-develop/pull/4345 into `phpunit/directives/wp-directive-processor.php` and ran unit tests. That added test passed (which it didn't when I un-applied your patch).

dmsnell: > it strikes me as odd that we need to remove a get_updated_html() call

yeah I've been working on figuring this out better

> still failing with your patch applied

this is why I had tried to carefully word that my patch fixed _my tests_ as I suspected we could still have more in this PR.

ockham: For the record, I've filed a [PR](https://github.com/WordPress/wordpress-develop/pull/4355) against `wordpress-develop` based on this one to make it a bit easier to reproduce and test fixes.

ockham: Noting that @dmsnell managed to reduce that test case to be more minimal and has filed a PR with a tentative fix: https://github.com/WordPress/wordpress-develop/pull/4371

ockham:
Here's an update and recap of the issue(s) with `wp-show`: https://github.com/WordPress/block-interactivity-experiments/issues/120#issuecomment-1519712031

ockham:
I've just removed a few unit tests that I introduced to help me pin-point the issue. The good news is that @dmsnell has filed a [PR](https://github.com/WordPress/wordpress-develop/pull/4371) against `wordpress-develop` that will fix them 🎉 

While testing that PR, I've noticed that its effect on `wp-show` seems to be equivalent to https://github.com/WordPress/block-interactivity-experiments/pull/141/commits/aad67cc7290f7ff9ad8a1f7f717b6cee56ea8063. Furthermore, I noticed that it seems to be possible to fix the [remaining issue](https://github.com/WordPress/block-interactivity-experiments/pull/215#issuecomment-1513162325) that I encountered with [this commit](https://github.com/WordPress/block-interactivity-experiments/pull/141/commits/3dbff3f2696b9f2e00cacd5dafb9bd40ba298c29).

This means that I hope that I've managed to finally unblock this PR 🤞 So it's ready for another round of reviews!

gziolo:
Excellent work on this PR. All the unit tests included make catching up with the implementation details easier. What's the reason for removing `test_directive_wraps_complex_content_in_template_if_when_is_false` in https://github.com/WordPress/block-interactivity-experiments/pull/141/commits/02a4a5be284841e74259ab5d6409511fb7cb6a2c? I like how it illustrates the more complex usage of the directive (excluding `process_wp_class` part). 

Have you explored how nesting of the same directive would work in practice? Example:

```html
<div data-wp-show="context.myBlock.open">
    I should be shown!
    <div data-wp-show="context.myOtherBlock.open">I should be shown!</div>
    <div data-wp-show="!context.myOtherBlock.open">I should not be shown!</div>
</div>
```

vs

```html
<div data-wp-show="context.myBlock.open">
    I should not be shown!
    <div data-wp-show="context.myOtherBlock.open">I should be shown!</div>
    <div data-wp-show="!context.myOtherBlock.open">I should not be shown!</div>
</div>
```

ockham:
> Excellent work on this PR. All the unit tests included make catching up with the implementation details easier. 

Thank you! ❤️ 

> What's the reason for removing `test_directive_wraps_complex_content_in_template_if_when_is_false` in [02a4a5b](https://github.com/WordPress/block-interactivity-experiments/commit/02a4a5be284841e74259ab5d6409511fb7cb6a2c)? I like how it illustrates the more complex usage of the directive (excluding `process_wp_class` part).

It's currently failing -- it would require `tags->get_updated_html()` calls after each `process_wp_*`. The test is basically a simplified version of `process_wp_directives`, with only the `wp-show` and `wp-class` directives enabled. While it was helpful in tracking down the underlying issue (which Dennis is solving in https://github.com/WordPress/wordpress-develop/pull/4371), I didn't really want to keep it around since it includes a call to `process_wp_class` -- so it's not really a _unit_ test for `wp-show` after all 😬 

Note that the directives SSR code was also affected by that bug, but I was able to fix it (https://github.com/WordPress/block-interactivity-experiments/pull/141/commits/aad67cc7290f7ff9ad8a1f7f717b6cee56ea8063) by adding that `get_updated_html()` call to `process_wp_directives`. (However, that workaround doesn't of course fix the failing test -- the test would still require the explicit `tags->get_updated_html()` calls after each `process_wp_*`. Another reason why I didn't think it was worth keeping the test around.)

> Have you explored how nesting of the same directive would work in practice? Example: [...]

I haven't yet! I can look into that in the next couple of days 👍

ockham:
> Have you explored how nesting of the same directive would work in practice? Example:
> 
> ```
> <div data-wp-show="context.myBlock.open">
>     I should be shown!
>     <div data-wp-show="context.myOtherBlock.open">I should be shown!</div>
>     <div data-wp-show="!context.myOtherBlock.open">I should not be shown!</div>
> </div>
> ```
> 
> vs
> 
> ```
> <div data-wp-show="context.myBlock.open">
>     I should not be shown!
>     <div data-wp-show="context.myOtherBlock.open">I should be shown!</div>
>     <div data-wp-show="!context.myOtherBlock.open">I should not be shown!</div>
> </div>
> ```

Added in https://github.com/WordPress/block-interactivity-experiments/pull/141/commits/35d7c99753a05e4b8ec58ac22ace03fd8bf0431b. @gziolo LMK if this covers what you meant or if you'd like something more or different! 😊

phpunit/directives/attributes/wp-show.php
```diff
@@ -0,0 +1,176 @@
...
+
+		while ( $tags->next_tag( array( 'tag_closers' => 'visit' ) ) ) {
+			process_wp_show( $tags, $context );
+		}
+
+		$updated_markup = <<<END
+			<template data-wp-show="context.myBlock.open"><div >
+				I should be shown!
+				<div data-wp-show="context.myOtherBlock.open">I should be shown!</div>
+				<template data-wp-show="!context.myOtherBlock.open"><div >I should not be shown!</div></template>
```
gziolo, line 170: Should it convert `div` to `template` inside another `template`? That was the reason I wanted to see those tests, so they cover edge cases like that.

phpunit/directives/attributes/wp-show.php
```diff
@@ -0,0 +1,176 @@
...
+			</div>
+END;
+		$this->assertSame( $updated_markup, $tags->get_updated_html() );
+		$this->assertSame( $context_before->get_context(), $context->get_context(), 'data-wp-show directive changed context' );
+	}
+
+	public function test_nested_directives_within_wp_show_with_falsy_value() {
+		$markup = <<<END
+			<div data-wp-show="context.myBlock.open">
+				I should be shown!
```
gziolo: Nitpick: when using the falsy value, it should follow other tests and the text used:
```suggestion
				I should not be shown!
```

src/directives/class-wp-directive-processor.php
dmsnell: maybe we could call the argument `$tag_name` to communicate that better? also I think we discussed this, but the part about "in uppercase" shouldn't matter, right?

src/directives/class-wp-directive-processor.php
dmsnell: and it's fixed now!

phpunit/directives/attributes/wp-show.php
gziolo: The expected result needs to get updated, too. Otherwise the test fails.

phpunit/directives/attributes/wp-show.php
```diff
@@ -0,0 +1,176 @@
...
+
+		$tags = new WP_Directive_Processor( $markup );
+
+		while ( $tags->next_tag( array( 'tag_closers' => 'visit' ) ) ) {
+			process_wp_show( $tags, $context );
+		}
+
+		$updated_markup = <<<END
+			<template data-wp-show="context.myBlock.open"><div >
+				I should be shown!
```
gziolo: We need it now to make the test pass:
```suggestion
				I should not be shown!
```

ockham: 😅

gziolo (approved):
I don't know the codebase, so take my feedback with a grain of salt. I'm approving this PR to show that the current testing coverage proves that the basic functionality is fully functional. There is one edge case that I highlighted in https://github.com/WordPress/block-interactivity-experiments/pull/141#discussion_r1184807429, which might be problematic, but we can test it only with the nested templated handled on the front end. It requires input from someone more knowledgable in the area.

src/directives/wp-process-directives.php
```diff
@@ -68,6 +68,7 @@ function wp_process_directives( $tags, $prefix, $directives ) {
 		foreach ( $attributes as $attribute ) {
 			call_user_func( $directives[ $attribute ], $tags, $context );
 		}
+		$tags->get_updated_html();
```
ockham: Since WP 6.2.1 -- which includes [this fix](https://github.com/WordPress/wordpress-develop/pull/4371) -- has now been released, we might be able to remove this line now (i.e. revert 42ab641c9a4dc9fc37e136eb596fc3eb288dd074).

ockham: Seems to be working: https://github.com/WordPress/block-interactivity-experiments/pull/141/commits/9c95d2df881051447818730f0e46d1b432ea7ee2

luisherranz:

This task was added to the Roadmap and we'll open a new issue/discussion once it's time to start working on this again.

PR #142, opened 2023-01-27
Create constants file for `wp-` prefixes

DAreRodz:
## What

Move all `wp-` prefixes (or strings containing any of them) to a constants file.

## Why

To make it easier to change them if necessary (e.g., `directivePrefix` to `data-wp-`).

## How

Creating a `constants.js` file.

https://github.com/WordPress/block-hydration-experiments/blob/4990097fef3f3c674eee195a3ea4bb2fd460c7cd/src/runtime/constants.js#L1-L3

I defined separated prefixes for components and directives, just in case we finally go for HTML-compliant directives (i.e., using [custom data attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*)).

### DIFF
--- a/src/runtime/vdom.js
+++ b/src/runtime/vdom.js
@@ -1,4 +1,9 @@
 import { h } from 'preact';
+import { directivePrefix as p } from './constants';
+
+const ignoreAttr = `${p}ignore`;
+const islandAttr = `${p}island`;
+const directiveParser = new RegExp(`${p}([^:]+):?(.*)$`);
 
 export const hydratedIslands = new WeakSet();
 
@@ -6,8 +11,8 @@ export const hydratedIslands = new WeakSet();
 export function toVdom(node) {
 	const props = {};
 	const { attributes, childNodes } = node;
-	const wpDirectives = {};
-	let hasWpDirectives = false;
+	const directives = {};
+	let hasDirectives = false;
 	let ignore = false;
 	let island = false;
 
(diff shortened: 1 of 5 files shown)


src/runtime/vdom.js
```diff
@@ -19,20 +24,20 @@ export function toVdom(node) {
 
 	for (let i = 0; i < attributes.length; i++) {
 		const n = attributes[i].name;
-		if (n[0] === 'w' && n[1] === 'p' && n[2] === '-' && n[3]) {
-			if (n === 'wp-ignore') {
+		if (n[p.length] && n.slice(0, p.length) === p) {
```
DAreRodz, line 27: I've changed how we check the attribute prefix. It turns out that using `String.prototype.slice` is much faster than the other approaches we tested [a while ago](https://github.com/WordPress/block-hydration-experiments/pull/99#issuecomment-1330956847).

The previous method that checked fixed positions is ~72% slower than the current one using `slice`. 😄 

https://jsbench.me/5lldeo3dr8/1

ockham (approved):
Code LGTM 👍 (I wasn't quite sure how to test though -- I take it we have some e2e coverage?)

DAreRodz:
> I wasn't quite sure how to test though -- I take it we have some e2e coverage?

We have e2e tests for full vDOM hydration, islands, and some directives. If none of them has broken, I guess these changes work. 🤞

src/runtime/vdom.js
luisherranz: You just blew my mind.

PR #143, opened 2023-02-01
Implement `wp-context` attribute directive

ockham:
Add the `wp-context` attribute directive (which we had previously only implemented as a _tag_ directive).

While handling the closing tag will require quite a bit of complex logic, I now think that that logic should go into the [`wp_process_directives` loop](https://github.com/WordPress/block-hydration-experiments/blob/bb6abaa5a4a7a01665b37544aca80636516a4431/wp-directives.php#L209-L245) rather than the individual directive processor.

### DIFF
--- a/phpunit/directives/attributes/wp-context.php
+++ b/phpunit/directives/attributes/wp-context.php
@@ -3,7 +3,7 @@
  * wp-context attribute directive test.
  */
 
-// require_once __DIR__ . '/../../../src/directives/attributes/wp-context.php'; // TODO
+require_once __DIR__ . '/../../../src/directives/attributes/wp-context.php';
 
 require_once __DIR__ . '/../../../src/directives/class-wp-directive-context.php';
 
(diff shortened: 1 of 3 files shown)


luisherranz (approved):
I don't know if the `$tags->is_tag_closer()` part is already available (I guess not, right?) but the code of the directive seems great to me. Pretty simple, actually 🙂

ockham:
Thank you for reviewing, @luisherranz!

> I don't know if the `$tags->is_tag_closer()` part is already available (I guess not, right?) but the code of the directive seems great to me. Pretty simple, actually 🙂

FWIW, `is_tag_closer()` is already [available in GB](https://github.com/WordPress/gutenberg/blob/1f47e41d4ce5987f9acf07fab58a7c39ea919c05/lib/compat/wordpress-6.2/html-api/class-wp-html-tag-processor.php#L1744-L1763) 😊

PR #144, opened 2023-02-01
SSR: Handle closing tags whose openers had an attribute directive

ockham:
Some attribute directives (such as `wp-context`, see #143), need to be run not only on an opening tag (e.g. `<div wp-context="...">` but also on the matching closing one. (For `wp-context`, this is required to reset the context to the value that corresponds to the enclosing scope's context.)

This needs somewhat complex logic to keep track of nested tags. Consider e.g. the following example (used in this PR's unit test):

```html
<div>Example: 
	<div foo-test="abc">
		<img>
		<span>This is a test></span>
		<div>Here is a nested div</div>
	</div>
</div>
```

Here, we want to run the `foo-test` attribute directive processor on the closing `</div>` that matches the opening `<div foo-test="abc">` -- but not on any of the other two closing `</div>`s.

Rather than requiring individual directive processors to implement the required tracking of nested tags, we implement it in the `wp_process_directives` function.

To that end, this PR also moves it into a separate file and changes its signature a bit for better testability. Finally, we're adding a unit test to verify the behavior described above.

### DIFF
--- a/phpunit/directives/wp-process-directives.php
+++ b/phpunit/directives/wp-process-directives.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * wp_process_directives test.
+ */
+
+require_once __DIR__ . '/../../src/directives/wp-process-directives.php';
+
+require_once __DIR__ . '/../../../gutenberg/lib/experimental/html/wp-html.php';
+
+class Helper_Class {
+	function process_foo_test( $tags, $context ) {
+	}
+}
+
+/**
+ * Tests for the wp_process_directives function.
+ *
+ * @group  directives
+ * @covers wp_process_directives
+ */
+class Tests_Directives_WpProcessDirectives extends WP_UnitTestCase {
+	public function test_correctly_call_attribute_directive_processor_on_closing_tag() {
+
+		// PHPUnit cannot stub functions, only classes.
+		$test_helper = $this->createMock( Helper_Class::class );
+
+		$test_helper->expects( $this->exactly( 2 ) )
+					->method( 'process_foo_test' )
+					->with(
+						$this->callback(
+							function( $p ) {
+								return 'DIV' === $p->get_tag() && (
+									// Either this is a closing tag...
+									$p->is_tag_closer() ||
+									// ...or it is an open tag, and has the directive attribute set.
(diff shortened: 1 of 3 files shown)


luisherranz (approved):
The code looks fine to me 🙂 

Although I thought the closing tag logic was going to be delegated to the HTML (Tag) Processor. Is it not the case anymore, or is this a temporary solution until [something like this](https://github.com/WordPress/gutenberg/pull/47573) is finished?

ockham:
> The code looks fine to me 🙂

Thank you! 🥳 

> Although I thought the closing tag logic was going to be delegated to the HTML (Tag) Processor. Is it not the case anymore, or is this a temporary solution until [something like this](https://github.com/WordPress/gutenberg/pull/47573) is finished?

I'm not sure at this point. We still need logic in the HTML (Tag) Processor to handle closing tags, in order to implement methods such as `set_content_within_balanced_tags` (which will also involve moving back and forth inside the HTML). (Note that OTOH, this current PR only ever moves forward.)

https://github.com/WordPress/gutenberg/pull/47573 has logic that's somewhat similar to what we're doing here (stack of opening tags). However, I didn't find it directly applicable to what we're doing here, and so I opted to implement it separately. I wouldn't rule out that eventually, we will have logic in the HTML (Tag) Processor that can serve both purposes, but they seemed different enough that I didn't see any harm in implementing this one separately already (rather than waiting for the HTML (Tag) Processor).

It's also possible that the purposes are different enough that we'll continue to handle things differently. (After all, we can optimize the logic here a bit to e.g only track opening tags of a given type 🤷‍♂️ )

cc/ @dmsnell 😊

dmsnell:
> `set_content_within_balanced_tags`

the thing that's been stewing lately and I hope to still get into soon is that I think we can avoid creating this and instead only expose the expected function, `set_inner_html()`.

this requires taking some time to understand the adoption agency algorithm, but from what I can tell it's mostly adding a few tag-type lists, maintaining a small tag stack (not unlike here and in our other HTML processing PRs), and crossing off all the cases of format boundaries.

this isn't to speak to this PR; go crazy here in your own code. for the HTML API I think that given that it does actually seem possible to do things 100% I'm tempted to avoid introducing things that are _like_ the 100% but aren't.

PR #145, opened 2023-02-01
Rename all mentions to client-side transitions with client-side navigation

luisherranz:
## What

Rename all mentions to `client-side transitions` with `client-side navigation`.

## Why

Because the term `transition` may be confusing as it's also used to refer to small changes in the page or animations. We believe `navigation` is a better term to describe them.

## How

Just rename everything, including comments, variable names, etc.


DAreRodz (approved):
Everything OK 👌 I'm gonna merge it as well.

PR #147, opened 2023-02-03
Serialize store in PHP

DAreRodz:
## What

Adds store serialization and hydration.

## Why

The framework requires them. Developers should be able to send the initial state from the server as part of the SSR.

## How

- [x] Implementing a `WP_Directive_Store` class that contains the generated state.
- [x] Adding a `store( $data )` function to define the state on the block's render callback.
- [x] Adding a `evaluate( $path, $context )` function to access both the state and the context (like in JS)
- [x] Serializing and sending the store's content to the client in a `<script type="application/json">` tag.
- [x] Parsing the serialized store in the client and using it to initialize the store.
- [x] Rename `wpx` to `store`

### DIFF
--- a/e2e/store-tag.spec.ts
+++ b/e2e/store-tag.spec.ts
@@ -1,0 +1,54 @@
+import { join } from 'path';
+import { test, expect } from '@playwright/test';
+
+test.describe('store tag', () => {
+	test('hydrates when it is well defined', async ({ page }) => {
+		await page.goto('file://' + join(__dirname, 'store-tag-ok.html'));
+		const value = page.getByTestId('counter value');
+		const double = page.getByTestId('counter double');
+		const clicks = page.getByTestId('counter clicks');
+
+		await expect(value).toHaveText('3');
+		await expect(double).toHaveText('6');
+		await expect(clicks).toHaveText('0');
+
+		page.getByTestId('counter button').click();
+
+		await expect(value).toHaveText('4');
+		await expect(double).toHaveText('8');
+		await expect(clicks).toHaveText('1');
+	});
+
+	test('does not break the page when missing', async ({ page }) => {
+		await page.goto('file://' + join(__dirname, 'store-tag-missing.html'));
+
+		const clicks = page.getByTestId('counter clicks');
+		await expect(clicks).toHaveText('0');
+		page.getByTestId('counter button').click();
+		await expect(clicks).toHaveText('1');
+	});
+
+	test('does not break the page when corrupted', async ({ page }) => {
+		await page.goto(
+			'file://' + join(__dirname, 'store-tag-corrupted-json.html')
+		);
+
+		const clicks = page.getByTestId('counter clicks');
+		await expect(clicks).toHaveText('0');
+		page.getByTestId('counter button').click();
+		await expect(clicks).toHaveText('1');
+	});
+
+	test('does not break the page when it contains an invalid state', async ({
+		page,
+	}) => {
+		await page.goto(
+			'file://' + join(__dirname, 'store-tag-invalid-state.html')
+		);
+
+		const clicks = page.getByTestId('counter clicks');
+		await expect(clicks).toHaveText('0');
+		page.getByTestId('counter button').click();
+		await expect(clicks).toHaveText('1');
+	});
+});
--- a/phpunit/directives/utils/evaluate.php
+++ b/phpunit/directives/utils/evaluate.php
@@ -18,11 +18,11 @@ class Tests_Utils_Evaluate extends WP_UnitTestCase {
 			array(
 				'state' => array(
 					'core' => array(
-						'a' => 1,
-						'b' => 2,
+						'a'      => 1,
+						'b'      => 2,
 						'nested' => array(
-							'c' => 3
-						)
+							'c' => 3,
+						),
 					),
 				),
 			)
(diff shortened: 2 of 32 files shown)


phpunit/directives/utils/evaluate.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+class Tests_Utils_Evaluate extends WP_UnitTestCase {
+	public function test_evaluate_function_should_access_state() {
+		// Init a simple store.
+		store(
+			array(
+				'state' => array(
+					'core' => array(
+						'a'      => 1,
+						'b'      => 2,
```
ockham: How about throwing in a different type value (to make sure `evaluate` gets the type right)? 😄 

```suggestion
						'b'      => true,
```

(and then `assertTrue` further down).

phpunit/directives/utils/evaluate.php
```diff
@@ -0,0 +1,56 @@
...
+		$this->assertSame( 3, evaluate( 'state.core.nested.c' ) );
+	}
+	public function test_evaluate_function_should_access_passed_context() {
+		$context = array(
+			'core' => array(
+				'a'      => 1,
+				'b'      => 2,
+				'nested' => array(
+					'c' => 3,
+				),
```
ockham: Maybe use a value for `$context` that's different from `$state`?

src/directives/attributes/wp-bind.php
```diff
@@ -17,7 +17,7 @@ function process_wp_bind( $tags, $context ) {
 
 		// TODO: Properly parse $value.
```
ockham: Let's maybe remove the `TODO` already -- it doesn't make sense to have it in every directive processor's individual code 😄 

```suggestion
```

src/directives/attributes/wp-class.php
```diff
@@ -17,7 +17,7 @@ function process_wp_class( $tags, $context ) {
 
 		// TODO: Properly parse $value.
```
ockham: Same as https://github.com/WordPress/block-hydration-experiments/pull/147/files#r1101622012

```suggestion
```

src/directives/attributes/wp-style.php
```diff
@@ -17,7 +17,7 @@ function process_wp_style( $tags, $context ) {
 
 		// TODO: Properly parse $value.
```
ockham: Same as https://github.com/WordPress/block-hydration-experiments/pull/147/files#r1101622012

```suggestion
```

src/directives/utils.php
```diff
@@ -1,15 +1,26 @@
...
-		$result = $context;
-		foreach ( $path as $key ) {
-			$result = $result[ $key ];
+require_once  __DIR__ . '/class-wp-directive-store.php';
+
+function store( $data ) {
+	WP_Directive_Store::merge_data( $data );
+}
+
+function evaluate( string $path, array $context = array() ) {
```
ockham, line 10: Let's maybe add a TODO here.

```suggestion
// TODO: Implement evaluation of complex logical expressions.
function evaluate( string $path, array $context = array() ) {
```

wp-directives.php
```diff
@@ -235,11 +238,24 @@ function wp_process_directives( $block_content ) {
...
+add_filter(
+	'render_block',
+	'wp_process_directives',
+	10,
+	1
+);
+
+// TODO: check if priority 9 is enough.
+// TODO: check if `wp_footer` is the most appropriate hook.
+add_action( 'wp_footer', WP_Directive_Store::render, 9 );
```
ockham: Ah, don't we normally use a different notation for invoking a class method? 🤔 

```suggestion
add_action( 'wp_footer', array( 'WP_Directive_Store', 'render' ), 9 );
```

e2e/store-tag.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+		await page.goto('file://' + join(__dirname, 'store-tag-ok.html'));
+		const value = page.getByTestId('counter value');
+		const double = page.getByTestId('counter double');
+		const clicks = page.getByTestId('counter clicks');
+
+		await expect(value).toHaveText('3');
+		await expect(double).toHaveText('6');
+		await expect(clicks).toHaveText('0');
+
+		page.getByTestId('counter button').click();
```
ockham, line 15: Can we also add unit test coverage for this functionality in the future (to test things in a more isolated way)? 😊 

I.e. for `wp-on:click`, we'd likely want to pass a [mock function](https://jestjs.io/docs/mock-function-api) and would check if it's really been called.

Or for `wp-bind`, we'd probably want to update a `state` value manually, and verify that the bound attribute is changed accordingly.

phpunit/directives/wp-directive-store.php
```diff
@@ -0,0 +1,137 @@
...
+					'core' => array(
+						'a' => 'overwritten',
+					),
+				),
+			),
+			WP_Directive_Store::get_data()
+		);
+	}
+
+	public function test_store_should_be_correctly_render() {
```
ockham: Grammar nitpick 😊 

```suggestion
	public function test_store_should_be_correctly_rendered() {
```

ockham (approved):
I left a few minor notes, but LGTM in general! 🚢

wp-directives.php
DAreRodz: I was sure there was a syntax for that, but I couldn't remember. 😄

e2e/store-tag.spec.ts
DAreRodz: Agree. We could do so in a future PR (also for other directives/functionalities). ☝️

luisherranz:
Great work, David 🙂

---

For context here, we also need to merge the state that comes in the new page during client-side navigations. 

I've been talking with David and we think merging both states should be enough. @darerodz can you open a new issue for that, and another one to also merge contexts?

SantosGuillamot:
> For context here, we also need to merge the state that comes in the new page during client-side navigations.

FYI: I think we will face this issue in [the Movies Demo](https://github.com/c4rl0sbr4v0/wp-movies-demo). There will probably be interactive blocks that aren't on the homepage, and after navigating to a movie, they won't work as expected. I'm thinking of a Tabs block to toggle between images and videos, for example.

PR #149, opened 2023-02-14
Handle missing class attribute inside `wp-class`

DAreRodz:
## What

Fixes a bug while hydrating `wp-class` directives. It happens when the `class` attribute is not defined―something that occurs when no class is server-side rendered. E.g., the following HTML

```html
<div wp-class:foo="state.falseValue"></div>
<button wp-on:click="actions.toggleFalseValue">Toggle</button>
```

currently throws this error:

Screenshot 2023-02-14 at 18 51 25

Also, when the button is clicked, the application crashes.

## Why

Directives should take care of those attributes they affect.

In this case, developers shouldn't be forced to write an empty class to make the `wp-class` directive work as expected.

## How

Simply using an empty string when `element.props.class` is undefined.

### DIFF
--- a/e2e/directives-class.html
+++ b/e2e/directives-class.html
@@ -52,6 +52,11 @@
 			data-testid="can toggle class in the middle"
 		></div>
 
+		<div
+			wp-class:foo="state.falseValue"
+			data-testid="can toggle class when class attribute is missing"
+		></div>
+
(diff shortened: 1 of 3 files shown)


luisherranz (approved):
LGTM!

PR #150, opened 2023-02-14
Fix `wp_process_directives` output

DAreRodz:
## What

Fixes a couple of bugs while processing directives and generating the new HTML.

## Why

The HTML output is basically `$block_content`, unchanged. 😅 

## How

1. Returning the updated HTML from the `WP_HTML_Tag_Processor`.
2. Getting the directive type from the attribute (`type:name`) before looking for the directive function.
3. Using the already implemented `wp-context` attribute.

I guess we'll have to implement tests for `wp_process_directives` sooner than later. ☝️

### DIFF
--- a/wp-directives.php
+++ b/wp-directives.php
@@ -216,7 +216,7 @@ function wp_process_directives( $block_content ) {
 	);
 
 	$attribute_directives = array(
-		// 'wp-context' => 'process_wp_context_attribute', // TODO
+		'wp-context' => 'process_wp_context_attribute',
 		'wp-bind'  => 'process_wp_bind',
 		'wp-class' => 'process_wp_class',
 		'wp-style' => 'process_wp_style',
@@ -234,12 +234,13 @@ function wp_process_directives( $block_content ) {
 			$attributes = $tags->get_attribute_names_with_prefix( 'wp-' );
 
 			foreach ( $attributes as $attribute ) {
-				if ( ! array_key_exists( $attribute, $attribute_directives ) ) {
+				list( $type ) = explode( ':', $attribute );
+				if ( ! array_key_exists( $type, $attribute_directives ) ) {
 					continue;
 				}
 
 				call_user_func(
-					$attribute_directives[ $attribute ],
+					$attribute_directives[ $type ],
 					$tags,
 					$context
 				);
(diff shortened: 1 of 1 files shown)


ockham:
Good spot, thank you!

Note that #144 changes `wp_process_directives` a bit: Rather than an HTML string, it will accept a `WP_HTML_Tag_Processor` object (typically created as `$tags = WP_HTML_Tag_Processor( $html );`), and return the modified `$tags`. This means that in order to plug it into the `render_block` filter (or anything that handles HTML strings really), we have to add a layer of indirection where that `$tags` object is created (also see #144).

This means that #144 (which has approval from @luisherranz 🎉 ) will conflict with this PR. There's another change in #144 related to the aforementioned indirection that might come in handy for Woo: `wp_process_directives` now also accepts separate arguments for prefix and lists of tag and attribute directives -- meaning it will be fully agnostic of those.

LMK if it's okay to merge #144 (I can help you carry over any necessary changes to Woo), or if you'd like to proceed differently 😊

DAreRodz:
> LMK if it's okay to merge https://github.com/WordPress/block-hydration-experiments/pull/144

@ockham, sure! We can merge #144 and adapt later this PR's changes (or even close it if it makes more sense).

ockham:
Great, thank you for confirming @DAreRodz !

ockham:
Merged #144.

> I guess we'll have to implement tests for `wp_process_directives` sooner than later. ☝️

I forgot to say, #144 also added [basic coverage](https://github.com/WordPress/block-hydration-experiments/blob/b3df431c3436bb5f41fede36b9307cd5d770dd67/phpunit/directives/wp-process-directives.php) for that function.

DAreRodz:
Great. 😊 @ockham, I'm closing this PR as most of the fixes were already addressed in your PR.

I think the part about getting the directive processors is still broken. My guess is that, if directive attributes have a colon (e.g., `wp-bind:value`), they won't be found inside `$attribute_directives`.

https://github.com/WordPress/block-hydration-experiments/blob/b3df431c3436bb5f41fede36b9307cd5d770dd67/src/directives/wp-process-directives.php#L33-L34

I'll take a look and open a new PR if necessary. 🕵️

DAreRodz:
> I'll take a look and open a new PR if necessary. 🕵️

My guess was right. 😅  I created a bugfix here: https://github.com/WordPress/block-hydration-experiments/pull/151

PR #151, opened 2023-02-16
Process directive attributes with a colon (e.g. `wp-bind:value`)

DAreRodz:
## What

The title is self-explanatory. 🙂 

## Why

Right now, directives using a colon are not processed and thus ignored because they're not found inside `$attribute_directives`:

https://github.com/WordPress/block-hydration-experiments/blob/b3df431c3436bb5f41fede36b9307cd5d770dd67/wp-directives.php#L219-L224

## How

Simply filtering the colon and the part that follows from attributes before doing the search.

### DIFF
--- a/phpunit/directives/wp-process-directives.php
+++ b/phpunit/directives/wp-process-directives.php
@@ -47,4 +47,18 @@ class Tests_Directives_WpProcessDirectives extends WP_UnitTestCase {
 		$tags   = new WP_HTML_Tag_Processor( $markup );
 		wp_process_directives( $tags, 'foo-', array(), $attribute_directives );
 	}
+
+	public function test_directives_with_type_and_name_are_processed_correctly() {
+		$test_helper = $this->createMock( Helper_Class::class );
+		$test_helper->expects( $this->atLeastOnce() )
+					->method( 'process_foo_test' );
+
+		$attribute_directives = array(
+			'foo-test' => array( $test_helper, 'process_foo_test' ),
+		);
+
+		$markup = '<div foo-test:value="abc"></div>';
+		$tags   = new WP_HTML_Tag_Processor( $markup );
+		wp_process_directives( $tags, 'foo-', array(), $attribute_directives );
+	}
 }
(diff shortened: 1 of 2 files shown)


phpunit/directives/wp-process-directives.php
```diff
@@ -47,4 +47,18 @@ function( $p ) {
 		$tags   = new WP_HTML_Tag_Processor( $markup );
 		wp_process_directives( $tags, 'foo-', array(), $attribute_directives );
 	}
+
+	public function test_directives_with_type_and_name_are_processed_correctly() {
```
DAreRodz: I called `type` and `name` the parts before and after the colon, respectively. Let me know if you come up with a better naming!

ockham: I can't right now 🤔 I think it's fine to merge as-is, we can change later.

We could try avoiding naming things altogether by calling the test something like

```suggestion
	public function test_directives_with_colon_processed_correctly() {
```

😬

src/directives/wp-process-directives.php
```diff
@@ -30,7 +30,14 @@ function wp_process_directives( $tags, $prefix, $tag_directives, $attribute_dire
 					}
 				}
 			} else {
+				// Helper that removes the part after the colon before looking
+				// for the directive processor inside `$attribute_directives`.
+				$get_directive_type = function ( $attr ) {
+					return strtok( $attr, ':' );
+				};
+
```
ockham, line 38: Should we maybe move this outside `wp_process_directives`'s function definition so we don't re-define the `$get_directive_type` ever so often?

ockham (approved):
Nice find! I think I probably messed something up in #144 with regard to this kind of directive. 

Great to have it fixed, and covered by a unit test 🎉  Also confirming that the test fails if I don’t apply the fix ✅  Thank you very much!

I left two minor suggestions -- otherwise LGTM!

src/directives/wp-process-directives.php
DAreRodz: Would it be fine if we create a global function just for that? That feels weird, haha. 😅 

Anyway, we can do it in another PR if you think is worth it. 👍

DAreRodz:
Thanks for your review and feedback, @ockham! 🙏

Issue #152, opened 2023-02-17
Remove directive tags (components) from the proposal


luisherranz:
I'd like to propose removing the directives tags (also referred to as components) from the code and the proposal.

The reasons are:

- During these months, we haven't been able to find a single use case, even a remote one, where using a directive tag is superior to using a directive attribute.
- If we publish the directive tags as they are intended today, we may lose the opportunity to use them in the future for other uses.

SantosGuillamot:
Thanks for opening this discussion 🙂 It makes sense to me! If they don't add value, the fewer concepts we add, the better. If we remove directive tags, maybe we can just talk about directives, instead of directive attributes. This way, it should be easier to explain and understand.

luisherranz:
> maybe we can just talk about directives, instead of directive attributes

I'd also love to hear opinions about this.

DAreRodz:
I see value in using directive tags for components that are only rendered in the browser (e.g., a hypothetical `<wp-graph type="column">` that renders a complex graph component and doesn't need server-side rendering).

Having said so, you can certainly use a "directive attribute" for that as well (e.g., `<div wp-graph="column">`) with the only consideration that you would probably have to pass custom props using `data-*` attributes.

> maybe we can just talk about directives, instead of directive attributes

If we follow that path and later add the directive tags again, I think we'll not be able to name them "directives". 😅

ockham:
I'm all for removing 👍 For some directives, components/directive tags don't make sense at all (e.g. `wp-bind` and friends).

> I see value in using directive tags for components that are only rendered in the browser (e.g., a hypothetical `<wp-graph type="column">` that renders a complex graph component and doesn't need server-side rendering).

My 2 cents: YAGNI 😬 If it's only hypothetical, we don't necessarily need to accommodate for it yet; we can (re-)introduce the concept of a tag directive if/when we feel we actually need it.

> maybe we can just talk about directives, instead of directive attributes

I'm all for that.

> If we follow that path and later add the directive tags again, I think we'll not be able to name them "directives". 😅

I wouldn't rule that out! We could later introduce them saying that in addition to the previous directives, which were all attributes, we're adding directives that are tags/components themselves; to differentiate them, we'll refer to them as "tag directives", and to the "other" ones as "attribute directives". I think it's sometimes necessary to give an existing notion new meaning; that's fine, as long as we differentiate clearly.

luisherranz:
Thanks, folks. 

[...]

PR #156, opened 2023-02-20
Add `wp-show` directive attribute

SantosGuillamot:
This pull request aims to replicate the behavior of the [`wp-show` directive tag (component)](https://github.com/WordPress/block-hydration-experiments/blob/main-wp-directives-plugin/src/runtime/components.js#L13-L22), but using a directive attribute instead.

It's true that in terms of functionality, the directive tag should be enough for now, but both approaches should be available. Especially if we end up [removing directive tags (components) from the proposal](https://github.com/WordPress/block-hydration-experiments/issues/152).

In [the movies demo](https://github.com/c4rl0sbr4v0/wp-movies-demo/), we will use this directive, so it would be great if we could just use directive tags.

### DIFF
--- a/e2e/directives-show.html
+++ b/e2e/directives-show.html
@@ -34,7 +34,20 @@
 			<p>falseValue children</p>
 		</div>
 
-		<div>Add block with context</div>
+		<div wp-context='{ "falseValue": false }'>
+			<div
+				wp-show="context.falseValue"
+				data-testid="can use context values"
+			>
+				falseValue
+			</div>
+			<button
+				wp-on:click="actions.toggleContextFalseValue"
+				data-testid="toggle context false value"
+			>
+				Toggle context falseValue
+			</button>
+		</div>
 
 		<script src="../build/e2e/store.js"></script>
 		<script src="../build/runtime.js"></script>
--- a/e2e/directives-show.html
+++ b/e2e/directives-show.html
@@ -1,0 +1,43 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<title>Directives -- wp-show</title>
+		<meta itemprop="wp-client-side-navigation" content="active" />
+	</head>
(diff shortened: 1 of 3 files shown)


luisherranz (changes requested):
The code looks fine to me, but would you mind adding some e2e tests? Thanks!! 🙏

SantosGuillamot:
> The code looks fine to me, but would you mind adding some e2e tests? Thanks!! 🙏

Sure, I'll do that 🙂

SantosGuillamot:
I started adding some tests, but something is not working as expected. It seems the HTML generated is not correct, and I don't know why it seems to be passing the attributes to the next HTML tag. I'll review tomorrow in more detail why that is happening.

SantosGuillamot:
I've added more tests to check the toggling and that it works with context as well -> [Commit](https://github.com/WordPress/block-hydration-experiments/pull/156/commits/374e810f73163cf052400bf6f74c656d36690444).

In [this other commit](https://github.com/WordPress/block-hydration-experiments/pull/156/commits/04d142a46c9d525f172229d3653ecab60f6e2810), @luisherranz fixed the issue I mention [here](https://github.com/WordPress/block-hydration-experiments/pull/156#issuecomment-1438807273). However, we have to decide [What should be the HTML structure returned when wp-show is false?](https://github.com/WordPress/block-hydration-experiments/issues/159) before merging this PR. The current approach is working, but if we go with the other alternative, [the problem mentioned](https://github.com/WordPress/block-hydration-experiments/pull/156#issuecomment-1438807273) still persists.

luisherranz:
Let's merge this as it is and if we decide to go with the opposite approach, we'll change it.

I'll investigate the bug a bit and open another issue, as it's not related to this PR (it's a general bug).

luisherranz (approved):

PR #157, opened 2023-02-20
SSR: Fix import paths, enable `wp-context` attribute directive processor

ockham:
Two groups of small fixes:

1. @DAreRodz pointed out to me that a few import paths were incorrect.
2. I'd forgotten to enable the `wp-context` attribute directive processor as part of #143.


DAreRodz (approved):
LGTM!

PR #158, opened 2023-02-21
WIP: Introduce directive processor

dmsnell:
The directive processor builds on the HTML Tag Processor to provide additional functionality, assuming normative markup generated from React, assuming that all tag directives contain a balanced closer.

This will break in practice, and is probably entirely wrong as-is. The purpose of this commit is to introduce the idea and explore using the code to get a feel for whether it will work.


src/class-wp-directive-processor.php
```diff
@@ -0,0 +1,92 @@
...
+				return true;
+			}
+
+			$depth--;
+		}
+
+		return false;
+	}
+
+	public function get_inner_html() {
```
adamziel: It's only `get_inner_html` if we assume normative markup. I'd suggest this:

```suggestion
	public function get_balanced_tags_html() {
```

Alternatively, we could return false in `next_balanced_closer` if it detects any misnesting.

src/class-wp-directive-processor.php
```diff
@@ -0,0 +1,92 @@
...
+		$start = $this->bookmarks['start']->end + 1;
+		$end   = $this->bookmarks['end']->start;
+
+		$this->release_bookmark( 'start' );
+		$this->release_bookmark( 'end' );
+
+		return substr( $this->html, $start, $end - $start );
+	}
+
+	public function set_inner_html( $new_html ) {
```
adamziel: Ditto 
```suggestion
	public function set_balanced_tags_html( $new_html ) {
```

src/class-wp-directive-processor.php
adamziel: Also, the current tag is a void element.

src/class-wp-directive-processor.php
```diff
@@ -0,0 +1,92 @@
...
+		return false;
+	}
+
+	public function get_inner_html() {
+		$this->set_bookmark( 'start' );
+		if ( ! $this->next_balanced_closer() ) {
+			$this->release_bookmark( 'start' );
+			return false;
+		}
+		$this->set_bookmark( 'end' );
```
adamziel: Bookmarking tag closers is affected by a bug, here's a fix:

https://github.com/WordPress/wordpress-develop/pull/4115

adamziel:
> This will break in practice, and is probably entirely wrong as-is

@dmsnell I left a few comments but I don't see anything fundamentally wrong here. The first draft is strong.

src/class-wp-directive-processor.php
dmsnell: thanks @adamziel - while not in the comments (but found in the commit message), because I was tossing this together quite quickly, I was taking this file to assume certain things:

> assuming normative markup generated from React, assuming that all tag directives contain a balanced closer.

so my idea was simply to write it as if that were true and we can treat deviations as bugs. I don't think any of the directives are self-closing foreign elements, though if they are we can add that support.

this is more or less my exploration of a client API where I consider strictness needs to be lower overall. e.g. if we're looking for the inner HTML of a `wp-context` then it won't matter if we have void tags or unbalanced tags. everything inside of the directives is blank since we're jumping straight to the next `wp-context` tag.

src/class-wp-directive-processor.php
dmsnell: excellent catch and excellent fix! I didn't test any of the code I wrote here, or even check if it's statically correct.

src/class-wp-directive-processor.php
adamziel: If the markup is supposed to be normative then there should be no harm in baling out in case misnesting is detected?

adamziel: But then again, assuming everything inside a directive is skippable like text means there’s no need to do these checks

dmsnell: we can't assume the contents are Text/Data. in fact, we _expect_ them to be HTML, which means we still need to scan them to avoid the conflation of things like `<wp-show><textarea>Use <wp-show when="…"> to conditionally show things</textarea></wp-show>`

src/directives/class-wp-directive-processor.php
```diff
@@ -0,0 +1,92 @@
...
+
+		return false;
+	}
+
+	public function next_balanced_closer() {
+		$depth = 0;
+
+		$tag_name = $this->get_tag();
+		while ( $this->next_tag( array( 'tag_name' => $tag_name, 'tag_closers' => 'visit' ) ) ) {
+			if ( ! $this->is_tag_closer() ) {
```
adamziel, line 43: This needs to check for void tags

src/directives/class-wp-directive-processor.php
```diff
@@ -0,0 +1,92 @@
...
+			}
+
+			$depth--;
+		}
+
+		return false;
+	}
+
+	public function get_inner_html() {
+		$this->set_bookmark( 'start' );
```
adamziel, line 59: `set_bookmark` may fail and that name may already be taken. I had the same issue in the HTML Processor. We need some way to set internal bookmarks in a different namespace.

src/directives/class-wp-directive-processor.php
```diff
@@ -0,0 +1,92 @@
...
+		if ( ! $this->next_balanced_closer() ) {
+			$this->release_bookmark( 'start' );
```
adamziel, line 61: How about wrapping this entire function in `try { } finally { $this->release_bookmark('start'); } `?

src/directives/class-wp-directive-processor.php
```diff
@@ -0,0 +1,92 @@
...
+		$depth = 0;
+
+		$tag_name = $this->get_tag();
+		while ( $this->next_tag( array( 'tag_name' => $tag_name, 'tag_closers' => 'visit' ) ) ) {
+			if ( ! $this->is_tag_closer() ) {
+				$depth++;
+				continue;
+			}
+
+			if ( 0 === $depth ) {
```
adamziel, line 48: Return `false` or throw if the tag name doesn't match?

src/directives/class-wp-directive-processor.php
```diff
@@ -0,0 +1,92 @@
...
+				return true;
+			}
+
+			$depth--;
+		}
+
+		return false;
+	}
+
+	public function get_inner_html() {
```
adamziel, line 58: `set_outer_html` will be challenging as we'll have to seek to an index that we'll also delete. In the HTML processor I ended up inventing "pinned" bookmarks that don't get updated or released when the lexical updates are applied.

adamziel: Actually scratch that, I tried `seek()`ing back to the original tag. That's not what you do here – I like it. Moving to the next tag makes things easier – I took a stab in [51bf340](https://github.com/WordPress/block-hydration-experiments/pull/158/commits/51bf3403a5f2d1ea628854f838dd3ce6f300e4a9).

src/directives/class-wp-directive-processor.php
dmsnell: thank you. this was not intended to be a correct working implementation so there are definitely more problems with it.

src/directives/class-wp-directive-processor.php
dmsnell: correct, though this was not intended to be a correct working implementation. I wanted to keep things terse to explore the idea and not distract the main functionality with all the details

src/directives/class-wp-directive-processor.php
dmsnell: I'm not very keen on using error-handling constructs like this as code conveniences. there's an obvious question it raises, "why is the `try` there?"

adamziel: because `finally { }` wouldn't work on its own 🤣

ockham:
I was trying to rebase #141 on this one and noticed it wasn't quite working for me. I added a bunch of unit test coverage, mostly taken from previous GB PRs that were going to implement `set_content_between_balanced_tags` to track down the issue.

The reason seems to be that we were missing some stuff in the GB compat layer (https://github.com/WordPress/gutenberg/pull/48681 and https://github.com/WordPress/gutenberg/pull/48692). With those PRs now merged and cherry-picked, the issue should be fixed by the release on GB 15.3 which I think should happen in the next couple of days.

dmsnell:
I'm stepping away from this PR, kinda regret posting the code now. it was supposed to be a kind of playground to explore interface ideas and now it seems more coupled to running code. I'd give it away to one of you but I don't think I can, so maybe I'll just unsubscribe. please feel free to create a new PR using any of the commits as a new branch tip. let me know if you no longer rely on this and we can close it.

src/directives/class-wp-directive-processor.php
ockham: Note that we're currently limiting the loop to [only visit tags whose name _does_ match.](https://github.com/WordPress/block-interactivity-experiments/pull/158/files#diff-ad36045951e27010af027ae380350ae4b07b56a659a3127b40b7967b2308d5bcR42)

src/directives/class-wp-directive-processor.php
ockham: Carrying over my [comment](https://github.com/WordPress/block-interactivity-experiments/pull/169#discussion_r1135390156) from #169 

> Note that since we're only visiting tags with the same name as the tag we're starting from, I think this case needs to be handled by bailing before even starting that loop. I.e. if we start at a `<br>`, the concept of a "matching closing tag" doesn't make sense, so we bail right away. If OTOH we start at a `<div>`, then we'll be only visiting other `<div>`s (and `</div>`s), so we don't need to check if we're encountering a void element.

ockham:
> let me know if you no longer rely on this and we can close it.

I'll go ahead and close this @dmsnell now that https://github.com/WordPress/block-interactivity-experiments/pull/169 has been merged -- hope that's okay! 😊

PR #161, opened 2023-02-27
Remove HTML Tag Processor Gutenberg requirement

SantosGuillamot:
If I am not mistaken, now that the Tag Processor is included in WordPress 6.2, we should remove the following dependency:

```php
require_once __DIR__ . '/../gutenberg/lib/experimental/html/wp-html.php';
```

After this Pull Request: [HTML API: Move into 6.2 Compat Folder since inclusion in Core](https://github.com/WordPress/gutenberg/pull/47749), that file doesn't exist anymore, and it seems to be causing issues. At least it's causing them locally to me.

Not 100% sure if this is the correct fix, you may know better than me @ockham .

### DIFF
--- a/wp-directives.php
+++ b/wp-directives.php
@@ -35,8 +35,6 @@ if ( ! is_plugin_active( 'gutenberg/gutenberg.php' ) ) {
 	return;
 }
 
-require_once __DIR__ . '/../gutenberg/lib/experimental/html/wp-html.php';
-
 require_once __DIR__ . '/src/directives/class-wp-directive-context.php';
 require_once __DIR__ . '/src/directives/class-wp-directive-store.php';
 require_once __DIR__ . '/src/directives/wp-process-directives.php';


SantosGuillamot:
> Not 100% sure if this is the correct fix.

Okay, I assume it is not the correct fix. Should we require the files from `gutenberg/lib/compat/wordpress-6.2/html-api/`?

ockham:
Hey @SantosGuillamot, thank you for filing this! I have a commit (e6a173ad4ba2cf5fb1f0da94456de61a3b16180f) that's actually tackling the imports (and also sort of centralize them, since they're currently kind of all over the place). I'll file a separate PR for that commit! 😄

ockham:
> I'll file a separate PR for that commit! 😄

https://github.com/WordPress/block-hydration-experiments/pull/162

ockham:
BTW -- just for clarification:

> If I am not mistaken, now that the Tag Processor is included in WordPress 6.2, we should remove the following dependency

The problem is that 6.2 hasn't been released yet 😬 So in order to be able to use the Tag Processor code, we'll need to continue to depend on the Gutenberg plugin at least until 6.2 is out. (However, it probably makes sense to depend on GB even longer, since we might want to use features that didn't make it into Core before the 6.2 Feature Freeze.)

SantosGuillamot:
Okay, that totally makes sense! Thanks for filing the PR 🙂 

I'm closing this one, then. Work continues on https://github.com/WordPress/block-hydration-experiments/pull/162

PR #162, opened 2023-02-27
Fix `WP_HTML_Tag_Processor` imports

ockham:
During preparation for the WP 6.2 release, the HTML Tag Processor files inside the Gutenberg plugin were [moved](https://github.com/WordPress/gutenberg/pull/47749) from `lib/experimental/html/` to `lib/compat/wordpress-6.2/html-api/`.

This PR updates the import locations accordingly, and sort of "centralizes" them, so we only reference the external dependency in one file (and can potentially later to point to e.g. `lib/compat/wordpress-6.3/html-api/class-gutenberg-html-tag-processor-6-3.php`, which has some features added after the WP 6.2 Feature Freeze, if needed).

See #161 for more context.

### DIFF
--- a/phpunit/directives/attributes/wp-bind.php
+++ b/phpunit/directives/attributes/wp-bind.php
@@ -7,7 +7,7 @@ require_once __DIR__ . '/../../../src/directives/attributes/wp-bind.php';
 
 require_once __DIR__ . '/../../../src/directives/class-wp-directive-context.php';
 
-require_once __DIR__ . '/../../../../gutenberg/lib/experimental/html/wp-html.php';
+require_once __DIR__ . '/../../../src/directives/wp-html.php';
 
 /**
  * Tests for the wp-bind directive.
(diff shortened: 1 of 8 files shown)


SantosGuillamot (changes requested):
It looks great to me, Bernie 🙂 

I guess [this requirement](https://github.com/WordPress/block-hydration-experiments/blob/fix/html-tag-processor-imports/wp-directives.php#L38) doesn't make sense anymore. Should we remove it or require `src/directives/wp-html.php` instead?

ockham:
> I guess [this requirement](https://github.com/WordPress/block-hydration-experiments/blob/fix/html-tag-processor-imports/wp-directives.php#L38) doesn't make sense anymore. Should we remove it or require `src/directives/wp-html.php` instead?

Ah, nice find! I missed that one, and that file isn't covered by unit tests 😬 

I've replaced it with `src/directives/wp-html.php` in e4ea115, since [we are creating a `WP_HTML_Tag_Processor` object](https://github.com/WordPress/block-hydration-experiments/blob/b095ffa74fa111f015095229203bfdd17390a860/wp-directives.php#L226) further down in that file 😄

SantosGuillamot (approved):

SantosGuillamot:
LGTM! 🙂

PR #163, opened 2023-02-27
Add client-side navigation support for `wp-ignore` directive

DAreRodz:
## What

Allows elements with the `wp-ignore` directive to keep their content after `navigate` calls.

## Why

During navigations, the content inside elements with `wp-ignore` could change, and those updates should not be reflected (those parts of the HTML document should be ignored).

## How

Implementing the `ignore` directive in a way that:
1. its content it's not parsed.
2. the value for the `dangerouslySetInnerHTML` prop never changes.

### DIFF
--- a/e2e/tovdom-full.spec.ts
+++ b/e2e/tovdom-full.spec.ts
@@ -3,10 +3,26 @@ import { test, expect } from '@playwright/test';
 
 test.describe('toVdom - full', () => {
 	test.beforeEach(async ({ page }) => {
-		await page.goto('file://' + join(__dirname, 'tovdom-full.html'));
+		await page.route('**/*.html', async (route, req) => {
+			const { pathname } = new URL(req.url());
+			route.fulfill({ path: join(__dirname, pathname) });
+		});
+
+		await page.goto('http://a.b/tovdom-full.html');
 	});
 
 	test('it should stop when it founds wp-ignore', async ({ page }) => {
+		const el = page.getByTestId('inside wp-ignore');
+		await expect(el).toBeVisible();
+	});
+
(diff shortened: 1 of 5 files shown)


e2e/tovdom-full.spec.ts
```diff
@@ -3,11 +3,38 @@ import { test, expect } from '@playwright/test';
...
+		await page.route('**/*.html', async (route, req) => {
+			const { pathname } = new URL(req.url());
+			route.fulfill({ path: join(__dirname, pathname) });
+		});
+		await page.route('**/*.js', async (route, req) => {
+			const { pathname } = new URL(req.url());
+			route.fulfill({ path: join(__dirname, '..', pathname) });
+		});
+
+		await page.goto('http://a.b/tovdom-full.html');
```
luisherranz, line 17: Would you mind applying this change to all the `spec` files to keep consistency among the tests?

luisherranz (approved):
I like it 🙂

e2e/tovdom-full.spec.ts
DAreRodz: No problem. 👍

DAreRodz: I've finally decided to do that in another PR using Playwright's fixtures instead of just copy-pasting that code to all spec files. The thing is that it would involve a bit of refactoring, which is out of the scope of this PR.

DAreRodz: @luisherranz I've created this PR: https://github.com/WordPress/block-hydration-experiments/pull/167

luisherranz: Perfect. Thank you David.

PR #164, opened 2023-02-28
Fetch scripts in the `<head>` on client-side navigations

### LINKED ISSUE #153: Fetch directives scripts after client-side navigation
In the same way we [support css in client-side navigation](https://github.com/WordPress/block-hydration-experiments/pull/85), new scripts should be fetched when navigating. Right now, if we navigate to a page with a new interactive block that wasn't present on the initial page, it may not work as expected, as the behavior defined in the store of that particular block won't be available.

For example, in [this movies demo](https://github.com/c4rl0sbr4v0/wp-movies-demo/), if we land on the homepage and navigate to a movie page, [the media tabs](https://github.com/c4rl0sbr4v0/wp-movies-demo/pull/19#issuecomment-1436547876) (a block only present in the movies pages) won't work because the actions needed haven't been defined.

michalczaplinski:
Fixes #153

This PR is a very crude first attempt to fetch the scripts on client-side navigations. It does not work with inline scripts, violates the `defer` and `async` attributes and probably breaks heaps of other things.

This functionality is however needed in order for the WP Movies Demo to showcase it's full capability so we should include (even a half-working) solution: 

> For example, in [this movies demo](https://github.com/c4rl0sbr4v0/wp-movies-demo/), if we land on the homepage and navigate to a movie page, https://github.com/c4rl0sbr4v0/wp-movies-demo/pull/19#issuecomment-1436547876 (a block only present in the movies pages) won't work because the actions needed haven't been defined.


src/runtime/router.js
```diff
@@ -105,6 +126,11 @@ export const init = async () => {
 		const body = toVdom(document.body);
 		hydrate(body, rootFragment);
 
+		// Cache the scripts. Has to be called before fetching the head.
```
luisherranz: Why is that?

src/runtime/router.js
```diff
@@ -105,6 +126,11 @@ export const init = async () => {
...
+		[].map.call(document.head.querySelectorAll('script'), (script) => {
```
luisherranz:
```suggestion
		[].map.call(document.head.querySelectorAll('script[src]'), (script) => {
```

Like the other one, right? 🙂

src/runtime/router.js
```diff
@@ -41,7 +42,27 @@ const fetchHead = async (head) => {
...
+					fetch(src).then((r) => r.text())
+				);
+			return scripts.get(src);
+		})
+	);
+
+	const scriptTags = scriptsToLoad.map((script) => {
+		const scriptEl = document.createElement('script');
+		scriptEl.textContent = script;
+		return scriptEl;
```
luisherranz: This doesn't execute the code until you add it to the DOM, right?

luisherranz (commented):
I think it's ok for a temporary fix 🙂

---

Right now it only works for the `head`, so maybe we should pass `document` to the `fetchHead` function and rename it to `fetchAssets` or something. It can still return the `head`, though.

```js
const fetchHead = async (head) => {
```

src/runtime/router.js
michalczaplinski: Yes, you're right!

src/runtime/router.js
michalczaplinski: Yes, that's right. Why do you ask? 🙂

src/runtime/router.js
michalczaplinski: I guess "has to" is a strong word. It should say "should be called". 

It's because if we don't cache the scripts that are present on the initial page load, then they will be fetched again inside of the `fetchHead()` because we call `fetchHead()` on the initial page load as well.

michalczaplinski: Now that I think about it, we should probably do the same for the styles 🙂

src/runtime/router.js
luisherranz: Out of curiosity. I wasn't 100% sure 😄

BTW, we would need to copy the `type="module"` here because those scripts are different.

src/runtime/router.js
luisherranz: Take into account that styles and scripts are not equal. Styles need to be added each time you navigate to a page because they may have been removed, but scripts only need to be executed once, even if they exist on multiple pages.

michalczaplinski: You're right, but in this case, I think it's okay. I'm only talking about a minor optimization: Add the stylesheets to the cache on the initial page load so that we don't have to download them again when we navigate to a new page. 

As you said, we still have to add _both_ the stylesheets currently present in the head and the "new" ones:

https://github.com/WordPress/block-hydration-experiments/blob/1d67b24047f4c7cd7692adc3468e642298a8acbe/src/runtime/router.js#L67-L68

But for scripts, we can only add the "new" ones: 

https://github.com/WordPress/block-hydration-experiments/blob/1d67b24047f4c7cd7692adc3468e642298a8acbe/src/runtime/router.js#L65

src/runtime/router.js
michalczaplinski: Absolutely 🙁

michalczaplinski: Done [0152fde](https://github.com/WordPress/block-hydration-experiments/pull/164/commits/0152fde6de644086fbd5614c05dff0e5830a0a66) 🙂

michalczaplinski:
Weird, the e2e tests seem to pass just fine locally:

Screenshot 2023-03-16 at 20 11 16

michalczaplinski:
Weird thing number 2: Locally there are 69 tests running, but in github actions there are 90 tests: 

https://github.com/WordPress/block-interactivity-experiments/actions/runs/4346748522/jobs/7800093352?pr=164#step:7:15

Possibly related to the number of workers? Locally, the tests run in 5 workers, whereas in CI, it's 1 worker.

SantosGuillamot:
I have just merged the main branch into this one. Now, I can see all the tests (96) locally, and the ones failing also fail locally.

SantosGuillamot:
I've been triaging the issue and I think we had two problems:

* The `fetchScriptOrStyle` function wasn't `async` and we were using `Promise.all(...).map`.
* We weren't passing `head` to the function, although we are using it.

I made a commit with what I considered a fix. Tests are passing now. Feel free to revert or modify it.

Apart from that, I've realized that we don't have `e2e` tests for this PR, nor for client-side navigation. I believe we should create some tests covering that:

* The content gets updated after client-side navigation.
* Styles are applied.
* New scripts work.

SantosGuillamot:
> Right now it only works for the head, so maybe we should pass document to the fetchHead function and rename it to fetchAssets or something. It can still return the head, though.

I believe we need this as well. Right now, in the [Movies demo](https://github.com/c4rl0sbr4v0/wp-movies-demo) for example, the `view.js` files are included in the `<body>` because they are needed after the `runtime`.

SantosGuillamot:
I've created initial tests to cover client-side navigation in [this pull request](https://github.com/WordPress/block-interactivity-experiments/pull/185).

EDIT: From what I can see there, the new scripts seem to work in the e2e tests when:
* We change `head` for `document` as @luisherranz suggested.
* We move [the script](https://github.com/WordPress/block-interactivity-experiments/blob/e2e-csn/e2e/html/csn-page-2.html#L30) to the `head`.

However, I am not able to make it work with the [Movies demo](https://github.com/c4rl0sbr4v0/wp-movies-demo).

michalczaplinski:
> However, I am not able to make it work with the [Movies demo](https://github.com/c4rl0sbr4v0/wp-movies-demo).

What do you mean by that? In what way did it not work?

SantosGuillamot:
> Right now it only works for the head, so maybe we should pass document to the fetchHead function and rename it to fetchAssets or something. It can still return the head, though.

I made [this commit](https://github.com/WordPress/block-interactivity-experiments/pull/164/commits/377cb1d9994ec1b8fbab6b376e58c133d09025c5) that aims to fetch the scripts in the whole document and not only the head.

It seems that this passes the [client-side navigation tests](https://github.com/WordPress/block-interactivity-experiments/pull/185), and I've checked it in the Movies demo, and it seems to work fine as well. Please take a look at the code because I am not sure the changes I made are 100% correct.

michalczaplinski:
I'm going to merge this now 🙁 Great work, Mario! 

My only concern is that I'd like us not to block when calling `fetchAssets()` in the future. Right now we always await it, for example [inside of the `init()`](https://github.com/WordPress/block-interactivity-experiments/blob/118bd8e27ef34542a983b9467abb9c3a56ae62a6/src/runtime/router.js#L161-L162) or when calling [`fetchPage()`](https://github.com/WordPress/block-interactivity-experiments/blob/118bd8e27ef34542a983b9467abb9c3a56ae62a6/src/runtime/router.js#L104-L105) but we should be able to delay `await`ing until the new `head` is retrieved from the `pages` cache, right before navigation to a new page happens.

I've also realized that we can simplify the code a little bit in the future because `querySelectorAll()` can match a group of elements like: `querySelectorAll('link[rel=stylesheet], script[type=module], script[src]')`.

PR #165, opened 2023-03-01
SSR: Remove `wp-context` tag directive

ockham:
Per https://github.com/WordPress/block-hydration-experiments/issues/152#issuecomment-1448482028, we've decided to remove tag directives, and only use attribute directives instead. 

As a consequence, this PR removes the only tag directive SSR processor we currently have: `wp-context`. It also changes the function signature of `wp_process_directives` plus its implementation to only cover attribute directives.

As a follow-up, we can consider moving the files found in `src/directives/attributes/` back into `src/directives/` (same for `phpunit/directives/attributes/`).


DAreRodz (approved):
LGTM! ✅ 

The only thing would be that we probably have to also change the name for the `attributes` folders, right? 🤷

ockham:
> LGTM! ✅
> 
> The only thing would be that we probably have to also change the name for the `attributes` folders, right? 🤷

Yeah, I'm even considering just moving everything from there to the parent `directives/` folder. (And maybe move the helper classes and functions to a separate dir.) But I wanted to keep that for a follow-up 😊

PR #168, opened 2023-03-03
Remove the `wp-context.php` import

michalczaplinski:
The import should have been removed in https://github.com/WordPress/block-hydration-experiments/pull/165

The e2e tests all pass even though this should clearly cause a server error. This is not a cause for panic of course 😅 but maybe we should consider running a full WP instance for the e2e tests in the future.


ockham (changes requested):
Oops, my bad! Thank you for catching this.

The `package-lock.json` seems a bit garbled; we might wanna remove that from the PR?

michalczaplinski:
> The package-lock.json seems a bit garbled; we might wanna remove that from the PR?

Ok, done 👍

ockham (approved):

PR #166, opened 2023-03-01
SSR: Try implementing `wp-each`

ockham:
_Currently using `add/directive-processor` from #158 as its base branch, in order to have access to the `WP_Directive_Processor` helper class, whose `set_inner_html` and `get_inner_html` methods this PR needs._

WIP. Based on a pair programming session with @luisherranz earlier today.

PR #169, opened 2023-03-06
SSR: Add basic directive processor

ockham:
WIP. This is basically a copy of @dmsnell's #158 for now, which I've forked per [this comment](https://github.com/WordPress/block-hydration-experiments/pull/158#issuecomment-1452908028).

I've removed @adamziel's `get_outer_html` and `set_outer_html` commit for now since we might not actually need it 😊


.wp-env.json
```diff
@@ -1,7 +1,7 @@
 {
 	"plugins": [
 		".",
-		"https://downloads.wordpress.org/plugin/gutenberg.latest-stable.zip"
+		"https://github.com/WordPress/gutenberg/releases/download/v15.3.0-rc.2/gutenberg.zip"
```
ockham: Temporary; revert once GB 15.3 has been released.

src/directives/class-wp-directive-processor.php
```diff
@@ -0,0 +1,92 @@
...
+			if ( ! $this->is_tag_closer() ) {
+				$depth++;
+				continue;
+			}
+
+			if ( 0 === $depth ) {
+				return true;
+			}
+
+			$depth--;
```
adamziel, line 38: I've left a few comments in the original PR from Dennis, they still apply here. For example, this loop needs to take into consideration void tags (like `<br>`) that don't really increase depth.

ockham: Thank you for the pointer!

> For example, this loop needs to take into consideration void tags (like `<br>`) that don't really increase depth.

Handled in 1a23ad9. Note that since we're only visiting tags with the same name as the tag we're starting from, I think this case needs to be handled by bailing before even starting that loop. I.e. if we start at a `<br>`, the concept of a "matching closing tag" doesn't make sense, so we bail right away. If OTOH we start at a `<div>`, then we'll be only visiting other `<div>`s (and `</div>`s), so we don't need to check if we're encountering a void element.

src/directives/class-wp-directive-processor.php
```diff
@@ -0,0 +1,155 @@
+<?php
+
+class WP_Directive_Processor extends WP_HTML_Tag_Processor {
```
ockham, line 3: #156 also had some more directive-specific logic (such as a `next_directive` method) that I removed. The reason is that it’ll need some more work where we’d need to move some of the logic that’s currently in `wp_process_directives` into this class. This isn’t currently a priority, so I’ve opted to defer it.

As a consequence however, `WP_Directive_Processor` doesn’t have any directive-specific code at all — it’s really more of a `WP_HTML_Tag_Processor`-derived class that allows finding a balanced closing tag, and getting/setting inner HTML. It might thus be worth considering renaming that class.

ockham:
Opening this for review. This is a blocker for a number of directives (#141, #166, #170), so it's about time to get the ball rolling 😊

luisherranz (approved):
This looks great to me, Bernie! 🎉

ockham:
Thanks you very much, Luis! 🎉

PR #170, opened 2023-03-06
SSR: Add `wp-html` and `wp-text` attribute directive processors

ockham:
Based on #169.

Per my understanding (and the [discussion below](https://github.com/WordPress/block-interactivity-experiments/pull/170#issuecomment-1456583117)), their only difference is that `wp-html` can contain HTML that is used unescaped, whereas `wp-text` escapes its value before rendering it.

### DIFF
--- a/wp-directives.php
+++ b/wp-directives.php
@@ -45,6 +45,7 @@ require_once __DIR__ . '/src/directives/attributes/wp-bind.php';
 require_once __DIR__ . '/src/directives/attributes/wp-context.php';
 require_once __DIR__ . '/src/directives/attributes/wp-class.php';
 require_once __DIR__ . '/src/directives/attributes/wp-style.php';
+require_once __DIR__ . '/src/directives/attributes/wp-text.php';
 
 function wp_directives_loader() {
 	// Load the Admin page.
@@ -217,6 +218,7 @@ function process_directives_in_block( $block_content ) {
 		'data-wp-bind'    => 'process_wp_bind',
 		'data-wp-class'   => 'process_wp_class',
 		'data-wp-style'   => 'process_wp_style',
+		'data-wp-text'    => 'process_wp_text',
 	);
 
 	$tags = new WP_HTML_Tag_Processor( $block_content );
--- a/wp-directives.php
+++ b/wp-directives.php
@@ -217,6 +217,7 @@ function process_directives_in_block( $block_content ) {
 		'data-wp-context' => 'process_wp_context',
 		'data-wp-bind'    => 'process_wp_bind',
 		'data-wp-class'   => 'process_wp_class',
+		'data-wp-html'    => 'process_wp_html',
 		'data-wp-style'   => 'process_wp_style',
 		'data-wp-text'    => 'process_wp_text',
 	);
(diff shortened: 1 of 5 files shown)


.wp-env.json
```diff
@@ -1,7 +1,7 @@
 {
 	"plugins": [
 		".",
-		"https://downloads.wordpress.org/plugin/gutenberg.latest-stable.zip"
+		"https://github.com/WordPress/gutenberg/releases/download/v15.3.0-rc.2/gutenberg.zip"
```
ockham: Temporary; revert once GB 15.3 has been released.

ockham:
@luisherranz Looking at https://github.com/WordPress/block-hydration-experiments/pull/118#issuecomment-1353407526, I'm wondering how (if?) `wp-text` and `wp-html` need to differ in terms of SSR.

One thing that comes to mind is escaping. Should `wp-html` run [`esc_html`](https://developer.wordpress.org/reference/functions/esc_html/) on the value it is passed? Are there any other differences? 🤔

luisherranz:
You can't escape the HTML of `wp-html`. This needs to work:

```php
<?php
wp_store( array( 
  'state' => array(
    'myPlugin' => array(
      'html' => get_some_raw_html_from_somewhere(),
    ),
  ),
) );
?>

<div data-wp-html="state.myPlugin.html">Default content</div>
```

I am not an expert on the matter, but `wp-text` should be the _safe_ one and `wp-html` the _unsafe_ one. We are basically using `dangerouslySetInnerHTML` underneath in the client.

By the way, should we be more explicit about the risks of `wp-html`? `wp-dangerous-html`? 😄

EDIT: I've added that question to the list of decisions that need to be made of the [Tracking Issue](https://github.com/WordPress/block-hydration-experiments/issues/80).

ockham:
> I am not an expert on the matter, but `wp-text` should be the _safe_ one and `wp-html` the _unsafe_ one. We are basically using `dangerouslySetInnerHTML` underneath in the client.

Ah, makes sense! 😅 So I guess I'll add escaping to `wp-text` then 🤔 

> By the way, should we be more explicit about the risks of `wp-html`? `wp-dangerous-html`? 😄

Maybe 🤔 Close enough to React, I guess.

ockham:
Opening this for review, now that #169 is in 😊

wp-directives.php
```diff
@@ -45,6 +45,7 @@ function () {
 require_once __DIR__ . '/src/directives/attributes/wp-context.php';
 require_once __DIR__ . '/src/directives/attributes/wp-class.php';
```
DAreRodz, line 46: I think the `require_once` for the `wp-html` directive is missing. 🤔 
```suggestion
require_once __DIR__ . '/src/directives/attributes/wp-class.php';
require_once __DIR__ . '/src/directives/attributes/wp-html.php';
```

DAreRodz: There is another `require_once` (line 38) that could cause some confusion, but it is not for the directive.

https://github.com/WordPress/block-interactivity-experiments/blob/ddec545967986e067997cd407c06aea8ca36afc9/wp-directives.php#L38

DAreRodz (approved):
Looks great to me! ✅ 

~Just adding one missing `require_once`, and it's ready to go. 😄~

Oh, nevermind, I didn't notice it was already addressed.

wp-directives.php
DAreRodz: Oh, nevermind. I just saw that it was already added. 👍

ockham:
Thank you @DAreRodz!

I'll rebase to change the prefix to `data-wp-` 😊

PR #173, opened 2023-03-08
Add `wp-text` directive

SantosGuillamot:
The goal of this pull request is to add the logic for the `wp-text` directive. It doesn't aim to cover the SSR as, if I am not mistaken, that would be handled in [this other PR](https://github.com/WordPress/block-hydration-experiments/pull/170).

While implementing it I wondered: Should it be possible to use more than one `wp-text` attribute in the same tag? If so, what would it happen? Would they join somehow or they are replaced? _I can move this to another discussion if you believe that's more appropriate_.

### DIFF
--- a/e2e/html/directives-text.html
+++ b/e2e/html/directives-text.html
@@ -1,0 +1,32 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<title>Directives -- wp-show</title>
+		<meta itemprop="wp-client-side-navigation" content="active" />
+	</head>
+	<body>
+		<div>
+			<span wp-text="state.text" data-testid="show state text"></span>
+			<button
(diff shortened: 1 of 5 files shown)


luisherranz:
> Should it be possible to use more than one wp-text attribute in the same tag?

I would not introduce more complexity for which we don't even know what use cases it would cover.

e2e/html/directives-text.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<title>Directives -- wp-show</title>
```
luisherranz:
```suggestion
		<title>Directives -- wp-text</title>
```

luisherranz (approved):
Perfect! Thanks, Mario 🚀

PR #174, opened 2023-03-09
Change naming from `block-hydration-experiments` to `block-interactivity-experiments`

SantosGuillamot:
As part of changing the name of the repo to `block-interactivity-experiments` instead of `block-hydration-experiments`, I'm creating this Pull Request to modify the mentions of it in the code.

Once this is done, I believe we can change safely change the name of the repo. According to [the docs](https://docs.github.com/en/repositories/creating-and-managing-repositories/renaming-a-repository):

* All existing information, with the exception of project site URLs, is automatically redirected to the new name.
* All `git clone`, `git fetch`, or `git push` operations targeting the previous location will continue to function. However, they strongly recommend updating it. Users can do this by using `git remote` on the command line:

```
$ git remote set-url origin NEW_URL
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,18 +1,18 @@
 {
-	"name": "wordpress/block-hydration-experiments",
+	"name": "wordpress/block-interactivity-experiments",
 	"type": "wordpress-plugin",
 	"license": "GPL-2.0-or-later",
-	"description": "Experiments with frontend hydration for blocks",
+	"description": "Experiments to add interactivity to blocks",
 	"homepage": "https://wordpress.github.io/gutenberg/",
 	"keywords": [
-		"block-hydration-experiments",
+		"block-interactivity-experiments",
 		"wordpress",
 		"wp",
 		"react",
 		"javascript"
 	],
 	"support": {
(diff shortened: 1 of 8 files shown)


composer.json
```diff
@@ -1,18 +1,18 @@
 {
-	"name": "wordpress/block-hydration-experiments",
+	"name": "wordpress/block-interactivity-experiments",
 	"type": "wordpress-plugin",
 	"license": "GPL-2.0-or-later",
-	"description": "Experiments with frontend hydration for blocks",
+	"description": "Experiments to add interactivity to blocks",
```
ockham: Maybe it makes sense to keep the word "frontend" here in some way?

```suggestion
	"description": "Experiments to add frontend interactivity to blocks",
```

or

```suggestion
	"description": "Experiments to add interactivity to blocks' frontends",
```

ockham (approved):
Thank you Mario! I left one minor suggestion -- otherwise, LGTM!

SantosGuillamot:
Thanks for the review and the suggestion, it totally makes sense! 🙂 I've just pushed a commit with that change.

PR #175, opened 2023-03-09
Create GitHub action to bump versions and upload plugin ZIP file

SantosGuillamot:
I'm creating a GitHub action that builds the WP Directive Plugin, bumps the version, and uploads it to the release. This way, it is easily downloadable directly from https://github.com/WordPress/block-hydration-experiments/releases/latest/download/block-interactivity-experiments.zip.

I'm doing the pull request on top of the `build-to-globals` branch because, while creating other plugins like the [WP Movies demo](https://github.com/c4rl0sbr4v0/wp-movies-demo/), they need to access functions like the `store`.

I've created a [fake 0.1.1 release](https://github.com/WordPress/block-hydration-experiments/releases) for testing purposes, and as a way to showcase that files are downloadable. We might need to delete it before merging this Pull Request.

### DIFF
--- a/.github/scripts/build-plugin.js
+++ b/.github/scripts/build-plugin.js
@@ -1,0 +1,16 @@
+const fs = require('fs');
+const archiver = require('archiver');
+
+// Create a new zip file
+const output = fs.createWriteStream('block-interactivity-experiments.zip');
+const archive = archiver('zip', { zlib: { level: 9 } });
+
+// Add files and directories to the zip file
+archive.directory('src/', 'src');
+archive.directory('build/', 'build');
+archive.file('Readme.md', { name: 'Readme.md' });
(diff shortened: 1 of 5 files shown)


luisherranz (approved):
I'm fine with whatever your decision is (and we can always change it later if needed), but just to make sure: this will bump the version (patch) and generate a new release each time a PR is merged, right?

EDIT: Nevermind, I just read it in [the movie's demo PR](https://github.com/c4rl0sbr4v0/wp-movies-demo/pull/20) 🙂

SantosGuillamot:
Thanks for the review!

> this will bump the version (patch) and generate a new release each time a PR is merged, right?

Yes, sorry I forgot to specify that here. I thought it wouldn't hurt to have rough versioning, and it could be useful to have different ZIP files depending on the release.

PR #176, opened 2023-03-09
Make `store` and `navigate` functions available through `window.__experimentalInteractivity`

SantosGuillamot:
Coded by @michalczaplinski. Not sure if you wanted to add anything to the branch, feel free to add/modify anything. I open the PR to be able to reference it better.

In order to be able to use some of the functions needed to create interactive blocks, we have to expose them somehow. In this pull request we are using `window.__experimentalInteractivity` as a workaround until we solve the bundling properly. With this, external plugins can access the store by doing:

```js
const { store } = window.__experimentalInteractivity;
```


luisherranz (approved):
This is fine to me. Thanks for using the `__experimental` prefix 🙂

SantosGuillamot:
In the [pull request where I added the GitHub action to create the plugin ZIP](https://github.com/WordPress/block-hydration-experiments/pull/175), I forgot we are not using `main` but `main-wp-directives-plugin`. As I merged the pull request on this branch, I just made a [commit](https://github.com/WordPress/block-hydration-experiments/pull/176/commits/efad3f9e51bb5b059c01e375c9f1dd6a4815cad2) here fixing that.

PR #177, opened 2023-03-09
Change plugins order in `wp-env` file to activate plugin automatically

SantosGuillamot:
Right now, if you run your local site using `wp-env`, the WP Directive plugin is not activated automatically. This happens because the plugin requires Gutenberg to be installed and we are installing the WP Directives plugin before Gutenberg. Just changing the order in the `.wp-env.json` file should be enough, if I am not mistaken.

### DIFF
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -1,7 +1,7 @@
 {
 	"plugins": [
-		".",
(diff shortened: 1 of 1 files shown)


luisherranz (approved):
👍

PR #181, opened 2023-03-16
Change directives prefix to `data-wp`

SantosGuillamot:
Based on the decision taken [here](https://github.com/WordPress/block-interactivity-experiments/issues/132#issuecomment-1466738441), I'm opening this pull request to handle it. So far I've:

- [x] Changed the constant to `data-wp`
- [x] Changed some filters that were adding directives to use `data-wp`
- [x] Changed the files handling the SSR to read from `data-wp`
- [x] Update the e2e tests.
- [x] Update the unit tests.

I still have to review properly if anything else is needed.

Some questions that came to my mind while working on it:

* Should we rename [the files handling the directives](https://github.com/WordPress/block-interactivity-experiments/tree/main-wp-directives-plugin/src/directives/attributes) in PHP as well? If so, what should they be named `data-wp-bind.php` or just `bind.php`?
* Should we change the [csnMetaTagItemProp](https://github.com/WordPress/block-interactivity-experiments/blob/main-wp-directives-plugin/src/runtime/constants.js#L1) to use `data-wp` as well? As it is not an attribute, I assumed it was correct without it, but let me know if you think different.


luisherranz (approved):
Test pass, so it looks good to me 🙂

By the way, I don't mind if we still refer to them as `wp-show`, even though in the HTML it will be used as `data-wp-show`.

michalczaplinski:
I found a couple of additional uses that we've missed:

Screenshot 2023-03-17 at 19 11 04

This also makes me realize that we've never removed the client components, even though we made a decision in https://github.com/WordPress/block-interactivity-experiments/issues/152#issuecomment-1448482028 to do so and we've removed the SSR of the `wp-context` component in https://github.com/WordPress/block-interactivity-experiments/pull/165

I ll make a new PR for this.

PR #182, opened 2023-03-16
Update paths to scripts in e2e tests

michalczaplinski:
It appears that the paths to scrips were incorrect. It's not clear to me why the e2e tests did not fail previously. 

Checking if this will fail in the github actions for some reason.


luisherranz:

PR #188, opened 2023-03-21
Fix `wp-text` directive in SSR

SantosGuillamot:
I was testing the recently added `wp-text` directive in SSR, and it doesn't seem to work in my case. I'm facing two different issues:

* Getting `Uncaught Error: Call to undefined method WP_HTML_Tag_Processor::set_inner_html()`.
* With JavaScript disabled, this is the HTML I'm getting:

Screenshot 2023-03-21 at 10 50 20

In order to solve that, I made a couple of commits, but I am not sure if it is the best way to solve it:
* In [the `process_directives_in_block` function](https://github.com/WordPress/block-interactivity-experiments/blob/main-wp-directives-plugin/wp-directives.php#L215-L229), I used the `WP_Directive_Processor` class instead of [`WP_HTML_Tag_Processor`](https://github.com/WordPress/block-interactivity-experiments/blob/main-wp-directives-plugin/wp-directives.php#L226), because is the one containing the `set_inner_html` function.
* I changed the end bookmark of the `set_inner_html` function to move one character to the left.

Please, let me know if this is correct and feel free to make any changes needed 🙂

### DIFF
--- a/src/directives/class-wp-directive-processor.php
+++ b/src/directives/class-wp-directive-processor.php
@@ -85,7 +85,7 @@ class WP_Directive_Processor extends WP_HTML_Tag_Processor {
 		list( $start_name, $end_name ) = $bookmarks;
 
 		$start = $this->bookmarks[ $start_name ]->end + 1;
-		$end   = $this->bookmarks[ $end_name ]->start - 1;
+		$end   = $this->bookmarks[ $end_name ]->start;
 
 		$this->seek( $start_name ); // Return to original position.
 		$this->release_bookmark( $start_name );
--- a/src/directives/class-wp-directive-processor.php
+++ b/src/directives/class-wp-directive-processor.php
@@ -85,7 +85,7 @@ class WP_Directive_Processor extends WP_HTML_Tag_Processor {
 		list( $start_name, $end_name ) = $bookmarks;
 
 		$start = $this->bookmarks[ $start_name ]->end + 1;
-		$end   = $this->bookmarks[ $end_name ]->start;
+		$end   = $this->bookmarks[ $end_name ]->start - 1;
 
 		$this->seek( $start_name ); // Return to original position.
 		$this->release_bookmark( $start_name );
--- a/wp-directives.php
+++ b/wp-directives.php
@@ -223,7 +223,7 @@ function process_directives_in_block( $block_content ) {
 		'data-wp-text'    => 'process_wp_text',
 	);
 
-	$tags = new WP_HTML_Tag_Processor( $block_content );
+	$tags = new WP_Directive_Processor( $block_content );
 	$tags = wp_process_directives( $tags, 'data-wp-', $directives );
 	return $tags->get_updated_html();
 }


SantosGuillamot:
Tests are failing, so it's probably not the correct solution 😄 Taking a look at it.

ockham:
> In [the `process_directives_in_block` function](https://github.com/WordPress/block-interactivity-experiments/blob/main-wp-directives-plugin/wp-directives.php#L215-L229), I used the `WP_Directive_Processor` class instead of [`WP_HTML_Tag_Processor`](https://github.com/WordPress/block-interactivity-experiments/blob/main-wp-directives-plugin/wp-directives.php#L226), because is the one containing the `set_inner_html` function.

That seems like the correct approach 🤔 

I'll take a deeper look in a bit!

src/directives/class-wp-directive-processor.php
```diff
@@ -85,7 +85,7 @@ public function set_inner_html( $new_html ) {
 		list( $start_name, $end_name ) = $bookmarks;
 
 		$start = $this->bookmarks[ $start_name ]->end + 1;
-		$end   = $this->bookmarks[ $end_name ]->start;
+		$end   = $this->bookmarks[ $end_name ]->start - 1;
```
ockham: Are you sure this is needed? I see that all failing tests now contain an extraneous character before the closing tag, e.g.

```
1) Tests_Directives_WpHtml::test_directive_sets_inner_html_based_on_attribute_value_and_retains_html
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'<div data-wp-html="context.myblock.someHtml">Lorem <em>ipsum</em> dolor sit.</div>'
+'<div data-wp-html="context.myblock.someHtml">Lorem <em>ipsum</em> dolor sit.></div>'
```

(Note the `>` right before the closing `</div>`.)

Maybe you've experienced different behavior when running the tests locally? You might want to destroy and recreate your local `wp-env`, as it's possible that you're using an outdated Gutenberg version. (There's been a [recent fix](https://github.com/WordPress/gutenberg/pull/48692) to adjust the end tag bookmarks -- if your local GB version doesn't have that yet, it would explain why you'd need the extra `- 1` there.)

SantosGuillamot:
I believe the one breaking the tests is the other commit:

* I changed the end bookmark of the set_inner_html function to move one character to the left.

wp-directives.php
```diff
@@ -223,7 +223,7 @@ function process_directives_in_block( $block_content ) {
 		'data-wp-text'    => 'process_wp_text',
 	);
 
-	$tags = new WP_HTML_Tag_Processor( $block_content );
+	$tags = new WP_Directive_Processor( $block_content );
```
ockham, line 226: Sorry I missed this!

(`wp-directives.php` is the one file that doesn't have unit test coverage, as it's kind of instantiating everything, so it'd be more of a case for integration testing.)

src/directives/class-wp-directive-processor.php
SantosGuillamot: Oh, okay. I'll take a look at that. I'm using GB 15.3.1. I'll take another look at it.

ockham: Hmm, I [cherry-picked it](https://github.com/WordPress/gutenberg/pull/48692#issuecomment-1452003538
) into the 15.3 branch though, so it should be there.

Anyway, if I locally revert ac35c9cfa0600ab0dffb8b1bfbe1c0bb3b60938e, tests pass for me. I can try to repro the `0/span>` issue...

SantosGuillamot: I've just tested with the `trunk` branch in Gutenberg, and it seems to work there. I've just reverted the commit. Thanks for taking a look at it, Bernie 🙂

ockham: Yay, happy to hear! 🎉

wp-directives.php
SantosGuillamot: No problem at all 🙂 We will need integration testing at some point that should handle this.

SantosGuillamot:
Tests are passing now 🙂 Do you believe it is ready to be merged @ockham ? Or is there anything missing?

ockham (approved):
LGTM! Thank you, Mario! :shipit:

PR #190, opened 2023-03-23
Rename PHP function to `wp_store` instead of `store`

SantosGuillamot:
As decided [here](https://github.com/WordPress/block-interactivity-experiments/issues/136#issuecomment-1463944790), I'm renaming the PHP function to `wp_store` instead of `store`.


michalczaplinski (approved):
Looks good to me!

michalczaplinski (approved):
Looks good to me!

PR #195, opened 2023-03-27
Add `.nvmrc` and `.editorconfig`

westonruter:
I tried doing `npm install` and I got the infamous:

```
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile 
npm WARN old lockfile This is a one-time fix-up, please be patient...
```

This should address that by adding an `.nvmrc` corresponding to `lockfileVersion:1`, assuming the user is using nvm. It's also what is in the Gutenberg repo.

This also copies the relevant declarations from the `.editorconfig` (also from Gutenberg), as I found my editor was by default adding spaces for indentation without it.


michalczaplinski (approved):
Looks good to me! 👍

PR #197, opened 2023-03-28
Add `replace` option to `navigate`

DAreRodz:
This PR adds a `replace` option to the router's `navigate` function.

When the option value is `false` (default), the function calls [`history.pushState`](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState).

When set to `true`, it calls [`history.replaceState`](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState) instead.

The idea of naming that option `replace` is based on what React Router does for its [`navigate`](https://reactrouter.com/en/main/hooks/use-navigate#type-declaration), or Next.js does for its [`Link`](https://nextjs.org/docs/api-reference/next/link#replace-the-url-instead-of-push) component.

### DIFF
--- a/e2e/page-2/store.js
+++ b/e2e/page-2/store.js
@@ -10,7 +10,6 @@ store({
 			state.newValue = !state.newValue;
 		},
 		replaceWithPage3: () => {
-			debugger;
 			navigate('/csn-page-3.html', { replace: true });
 		},
 	},
--- a/e2e/page-2/store.js
+++ b/e2e/page-2/store.js
@@ -1,4 +1,5 @@
 import { store } from '../../src/runtime/store';
+import { navigate } from '../../src/runtime/router';
 
 store({
 	state: {
@@ -7,6 +8,10 @@ store({
 	actions: {
 		toggleNewValue: ({ state }) => {
 			state.newValue = !state.newValue;
+		},
+		replaceWithPage3: () => {
+			debugger;
+			navigate('/csn-page-3.html', { replace: true });
 		},
 	},
 });
--- a/e2e/specs/csn.spec.ts
+++ b/e2e/specs/csn.spec.ts
@@ -50,4 +50,16 @@ test.describe('toVdom - full', () => {
 		await page.getByTestId('toggle newValue').click();
 		await expect(el).toBeHidden();
 	});
+
+	test('it should replace current page in session history when using `replace` option', async ({
+		page,
+	}) => {
+		await page.getByTestId('csn-next-page').click();
+		await page.getByTestId('replace with page 3').click();
+		const newTitle = page.getByTestId('csn-heading-page-3');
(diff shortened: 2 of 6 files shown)


src/runtime/router.js
```diff
@@ -115,14 +115,15 @@ export const prefetch = (url) => {
...
+export const navigate = async (href, { replace = false } = {}) => {
 	const url = cleanUrl(href);
 	prefetch(url);
 	const page = await pages.get(url);
 	if (page) {
 		document.head.replaceChildren(...page.head);
 		render(page.body, rootFragment);
-		window.history.pushState({}, '', href);
+		if (replace) window.history.replaceState({}, '', href);
+		else window.history.pushState({}, '', href);
```
luisherranz: Save 23B? 😬

```suggestion
		window.history[replace ? "replaceState" : "pushState"]({}, '', href);
```

e2e/page-2/store.js
```diff
@@ -8,5 +9,9 @@ store({
 		toggleNewValue: ({ state }) => {
 			state.newValue = !state.newValue;
 		},
+		replaceWithPage3: () => {
+			debugger;
```
luisherranz:
```suggestion
```

e2e/specs/csn.spec.ts
```diff
@@ -50,4 +50,16 @@ test.describe('toVdom - full', () => {
...
+	test('it should replace current page in session history when using `replace` option', async ({
+		page,
+	}) => {
+		await page.getByTestId('csn-next-page').click();
+		await page.getByTestId('replace with page 3').click();
+		const newTitle = page.getByTestId('csn-heading-page-3');
+		await expect(newTitle).toHaveText(/Client-side navigation Page 3/);
+		await page.goBack();
+		const prevTitle = page.getByTestId('csn-heading-page-1');
+		await expect(prevTitle).toHaveText(/Client-side navigation Page 1/);
```
luisherranz:
```suggestion
		// We start on page 1 and navigate using `push` to page 2.
		await page.getByTestId('csn-next-page').click();
		// Once we are in page 2, we navigate using `replace` to page 3.
		await page.getByTestId('replace with page 3').click();
		const newTitle = page.getByTestId('csn-heading-page-3');
		await expect(newTitle).toHaveText(/Client-side navigation Page 3/);
		// If we go back, we should go back to page 1 and not 2.
		await page.goBack();
		const prevTitle = page.getByTestId('csn-heading-page-1');
		await expect(prevTitle).toHaveText(/Client-side navigation Page 1/);
```

luisherranz (approved):
LGTM! Thanks David 🎉

e2e/page-2/store.js
DAreRodz: 🙈

PR #199, opened 2023-03-28
Remove type from prop inside `WP_Directive_Store`

DAreRodz:
[Typed properties](https://www.php.net/manual/en/migration74.new-features.php#migration74.new-features.core.typed-properties) inside classes is a feature introduced in PHP 7.4.

The latest WordPress version (6.1) supports PHP 5.6, so it'd be better to stop using typed properties.

### DIFF
--- a/src/directives/class-wp-directive-store.php
+++ b/src/directives/class-wp-directive-store.php
@@ -1,7 +1,7 @@
 <?php
 
 class WP_Directive_Store {
-	private static array $store = array();
+	private static $store = array();
 
 	static function get_data() {
 		return self::$store;


SantosGuillamot (approved):
LGTM 🙂 Thanks for the fix, David!

westonruter:
Perhaps [`PHPCompatibility`](https://github.com/PHPCompatibility/PHPCompatibility) sniffs should be added to check for this automatically?

westonruter:
Ah, it is added already. But `phpcs` is not being run in the checks.

westonruter:
Looks like there are two more instances to fix:

```
$ vendor/bin/phpcs --standard=PHPCompatibilityWP --extensions=php --runtime-set testVersion 5.6- *.php src phpunit

FILE: src/directives/utils.php
--------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------------
 10 | ERROR | 'string' type declaration is not present in PHP version 5.6 or earlier
--------------------------------------------------------------------------------------
```

https://github.com/WordPress/block-interactivity-experiments/blob/4674ba853e560492b47ec75a5d6601eea260b007/src/directives/utils.php#L10

```
FILE: phpunit/directives/wp-directive-store.php
-------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
-------------------------------------------------------------------------------------------------------
 134 | ERROR | Trailing comma's are not allowed in function calls in PHP 7.2 or earlier
-------------------------------------------------------------------------------------------------------
```
https://github.com/WordPress/block-interactivity-experiments/blob/4674ba853e560492b47ec75a5d6601eea260b007/phpunit/directives/wp-directive-store.php#L134

westonruter:
> Looks like there are two more instances to fix:

I've fixed these in https://github.com/WordPress/block-interactivity-experiments/pull/201 and cherry-picked your commit as well.

DAreRodz:
I see this PR will be closed once #201 is merged, which also contains changes from this PR. No further action is needed here. 🙂

PR #208, opened 2023-04-02
Implement `next_directive` in `WP_Directive_Processor`

ockham:
WIP.

TODO:
- [ ] Store bookmarks to open tags in `$this->tag_stack`.
- [ ] Add new method (or logic) to go to matching opener when on a closer.
- [ ] Replace calls to `next_tag` with `next_directive`.

This will allow directive processors (iterative or recursive) to easily locate the opener. We might want to split our current directive processors into an opener and an optional closer, e.g.

```
	$directives = array(
		'data-wp-context' => array( 'process_wp_context_opener', `process_wp_context_closer` ),
		'data-wp-bind'    => 'process_wp_bind',
		// etc
	);
```

We can then try to implement the recursive way.

### DIFF
--- a/phpunit/directives/wp-process-directives.php
+++ b/phpunit/directives/wp-process-directives.php
@@ -5,7 +5,7 @@
 
 require_once __DIR__ . '/../../src/directives/wp-process-directives.php';
 
-require_once __DIR__ . '/../../src/directives/wp-html.php';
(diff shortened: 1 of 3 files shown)


luisherranz:
Please, feel free to open a new discussion in the [Interactivity API category](https://github.com/WordPress/gutenberg/discussions/categories/interactivity-api) if you think this should be explored further.

PR #213, opened 2023-04-14
wp-bind improvements and negation in runtime

luisherranz:
## What

- I've added support for negations to the JavaScript runtime: `data-wp-bind.checked="!state.checked"`.
- I've improved `data-wp-bind` to modify attributes on hydration (Preact doesn't modify them when using `hydrate`).
- I've added e2e tests for `data-wp-bind`, including one that uses `"!state.checked"`.

## Why

Because `!` is easy enough to add, and it's very useful when combined with `data-wp-show` or `data-wp-bind.hidden` (if/else).

## How

- For the negation, just check if the path starts with `!`.
- For the `data-wp-bind`, add a `useEffect` and change the DOM manually to reflect the vDOM (similar to `data-wp-class`).

## Notes

1. While working on `data-wp-bind.checked`, I stumbled upon this weird thing: When you change the `checked` attribute of a node with React ([reproduced here](https://stackblitz.com/edit/vitejs-vite-yyynmq?file=index.html,src%2Fmain.jsx&terminal=dev)) or Preact ([reproduced here](https://stackblitz.com/edit/vitejs-vite-ph3ck3?file=src%2Fmain.jsx,index.html&terminal=dev)), the browser doesn't update the `checked` attribute of the component. Everything seems to work because `element.checked` is correct and [Dan says this is correct way of doing it](https://github.com/facebook/react/issues/24439), but at first it's pretty confusing 😅
2. I'll work on the server-side negation in another PR.

### DIFF
--- a/e2e/html/directive-bind.html
+++ b/e2e/html/directive-bind.html
@@ -1,0 +1,40 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<title>Directives -- data-wp-bind</title>
+		<meta itemprop="wp-client-side-navigation" content="active" />
+	</head>
+	<body>
+		<a
+			data-wp-bind.href="state.url"
+			data-testid="add missing href at hydration"
+		></a>
+
+		<a
+			href="/other-url"
+			data-wp-bind.href="state.url"
+			data-testid="change href at hydration"
+		></a>
+
+		<input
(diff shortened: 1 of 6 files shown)


SantosGuillamot (approved):
Looks great to me! 🙂

PR #214, opened 2023-04-14
Add server side negation operator

luisherranz:
## What

The server-side part of the `!` operator introduced in https://github.com/WordPress/block-interactivity-experiments/pull/213.

## Why

Because the server also needs to understand it.

## How

Just replicate the JavaScript logic, check for the presence of `!` at the start of the $path and return the opposite if it's found.


SantosGuillamot (approved):
I've tested it an it works as expected. Looks good to me 🙂

PR #216, opened 2023-04-18
Remove the router and leave only the Island hydration

DAreRodz:
This is a simple PR that removes anything related to the router and client-side navigation, as they don't need to be included in Gutenberg at the moment.

- [x] Removed all code in `router.js` except the one for island hydration
- [x] Refactored e2e tests to use `data-wp-island` instead of the CSN meta tag
- [x] Removed e2e tests for full hydration and client-side navigation
- [x] Removed `data-wp-link` directive
- [x] Renamed `router.js` to `hydration.js`


src/runtime/hydration.js
```diff
@@ -0,0 +1,15 @@
+import { hydrate } from 'preact';
+import { toVdom, hydratedIslands } from './vdom';
+import { createRootFragment } from './utils';
+import { directivePrefix } from './constants';
+
+// Initialize the router with the initial DOM.
```
luisherranz:
```suggestion
```

luisherranz (approved):
LGTM David! Feel free to merge.

PR #220, opened 2023-04-18
Remove `wp-show` and `wp-text` directives

DAreRodz:
This PR eliminates `wp-show` and `wp-text` directives, as they're not required for now. The `wp-link` directive, which is also not needed, has already been removed in #216.


e2e/specs/tovdom-islands.spec.ts
```diff
@@ -28,7 +28,7 @@ test.describe('toVdom - isands', () => {
 		await expect(el).toBeVisible();
 	});
 
-	test('directives inside islands should not be hydrated twice', async ({
+	test.skip('directives inside islands should not be hydrated twice', async ({
```
DAreRodz, line 33: I skipped this one because I didn't see clearly how to refactor this test without using `wp-show`. @luisherranz, can you take a look at this?

DAreRodz: We can remove this test for now, although I would prefer to have this case covered.

luisherranz: We can do something like this:

```html
<div data-wp-island>
  <div data-wp-island>
    <div data-testid="island inside another island">
      <div data-wp-init="actions.appendText">
        <span>
          This should not have two spans appended because that means we hydrated
          twice.
        </span>
      </div>
    </div>
  </div>
</div>
```

```js
appendText: ({ ref }) => {
  const span = document.createTextNode("Span");
  span.innerText = "some text";
  ref.appendChild(span);
};
```

And check that there's only one span or something similar.

But I think `ref` is currently broken due to [this bug](https://github.com/WordPress/block-interactivity-experiments/issues/191) and `data-wp-init` is not implemented yet, so we can wait until then.

e2e/html/tovdom-islands.html
```diff
@@ -5,14 +5,14 @@
 	</head>
 	<body>
 		<div>
-			<div data-wp-show="state.falseValue">
+			<div data-wp-bind.hidden="!state.falseValue">
```
luisherranz, line 8: There's a `state.trueValue` as well if you prefer.

luisherranz (approved):
LGTM, David!

PR #223, opened 2023-04-20
Remove _wrapped, expose directives and expose full store

luisherranz:
## What

- Remove the `_wrapper` prop.
- Keep directives in the DOM.
- Pass the whole store to evaluate.

## Why

- Remove the `_wrapper` prop: because sometimes it's leaked to the DOM.
- Keep directives in the DOM: to avoid confusion when inspecting the DOM while coding.
- Pass the whole store to evaluate (instead of just state): to be able to access other arbitrary store properties in the callbacks, like `selectors`.

## How

It's explained in this video.

<a href="https://www.loom.com/share/049ef0e25cf244d593a77466d04139cd">
    
  </a>

https://www.loom.com/share/049ef0e25cf244d593a77466d04139cd


DAreRodz (approved):
💯 ✅

PR #224, opened 2023-04-21
Fix `evaluate` so it is always the same function instance

### LINKED ISSUE #191: Refs are not passed correctly to directive callbacks
While experimenting with Luis, we noticed that the `ref` attribute the `evaluate` function injects to directive callbacks could be incorrectly passed.

It happens when `ref.current` is `null`.

In that case, a `null` reference is passed here to the `getEvaluate` function.

https://github.com/WordPress/block-interactivity-experiments/blob/4765ff57d21f8d4917ae8f5ae4c0043afa9def42/src/runtime/hooks.js#L47

That function then passes down the `null` reference to the `value` function.

https://github.com/WordPress/block-interactivity-experiments/blob/4765ff57d21f8d4917ae8f5ae4c0043afa9def42/src/runtime/hooks.js#L33-L39

We should pass the `ref` object to `getEvaluate` instead of `ref.current`, and pass `ref.current` to `value` as argument.

DAreRodz:
The `evaluate` function that directives receive was generated by passing the `ref.current` value to `getEvaluate`, instead of `ref` directly. That forced the `Directive` wrapper to generate a new instance of `evaluate` each time `ref.current` changed, a new instance not referenced [here](https://github.com/WordPress/block-interactivity-experiments/blob/efcfa2a038be88ed5a97acdf668b5047a719ad77/src/runtime/directives.js#L51) in `wp-effect`.

This PR fixes that.

Related: https://github.com/WordPress/block-interactivity-experiments/issues/191

### DIFF
--- a/src/runtime/hooks.js
+++ b/src/runtime/hooks.js
@@ -1,5 +1,5 @@
 import { h, options, createContext } from 'preact';
-import { useRef } from 'preact/hooks';
+import { useRef, useMemo } from 'preact/hooks';
 import { rawStore as store } from './store';
 
 // Main context.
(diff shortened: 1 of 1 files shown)


luisherranz (approved):
LGTM!

PR #225, opened 2023-04-21
Add wp-init and other small adjustments

luisherranz:
## What

Add `data-wp-init` and other small adjustments for this: https://github.com/WordPress/gutenberg/issues/49868

## Why

We need a basic version of `wp-init`, and there are some things that we won't need in Gutenberg.

## How

For now, just add the `wp-init` copying `wp-effect` but using `useEffect` with an empty dependency array instead of `useSignalEffect`. Also add `return`'s to be able to execute callbacks when the elements are removed.

I also removed the `DOMContentLoaded` event because we will use `defer` or `type="module"` in Gutenberg.

I also removed some unnecessary exports.

### DIFF
--- a/src/runtime/constants.js
+++ b/src/runtime/constants.js
@@ -1,2 +1,1 @@
-export const csnMetaTagItemprop = 'wp-client-side-navigation';
 export const directivePrefix = 'data-wp-';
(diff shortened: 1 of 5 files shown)


DAreRodz (approved):
LGTM! 👍

PR #226, opened 2023-04-21
Add a `useSignalEffect` implementation that behaves like `useEffect`

DAreRodz:
The current `useSignalEffect` hook implementation from [@preact/signals](https://github.com/preactjs/signals) calls the passed callback before changes to the DOM are made, but it should wait for DOM changes (like `useEffect` does). 

I just copy-pasted the `useSignalEffect` hook and its dependencies from https://github.com/preactjs/signals/pull/290 until those changes are merged and released in [@preact/signals](https://github.com/preactjs/signals).

### DIFF
--- a/src/runtime/directives.js
+++ b/src/runtime/directives.js
@@ -58,7 +58,7 @@ export default () => {
 		const contextValue = useContext(context);
 		Object.entries(on).forEach(([name, path]) => {
 			element.props[`on${name}`] = (event) => {
-				return evaluate(path, { event, context: contextValue });
+				evaluate(path, { event, context: contextValue });
 			};
 		});
 	});
--- a/src/runtime/directives.js
+++ b/src/runtime/directives.js
@@ -48,7 +48,17 @@ export default () => {
 		const contextValue = useContext(context);
 		Object.values(effect).forEach((path) => {
 			useSignalEffect(() => {
-				evaluate(path, { context: contextValue });
+				return evaluate(path, { context: contextValue });
+			});
+		});
+	});
+
+	// data-wp-init.[name]
+	directive('init', ({ directives: { init }, context, evaluate }) => {
+		const contextValue = useContext(context);
+		Object.values(init).forEach((path) => {
+			useEffect(() => {
+				return evaluate(path, { context: contextValue });
 			});
 		});
 	});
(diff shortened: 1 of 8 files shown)


luisherranz:
I've been trying to add tests to reproduce the problem [here](https://github.com/WordPress/gutenberg/pull/44509/files#diff-ac3ab677b419ffc1a9537b403c75d082df4ce82942495c2fd7b880e83844002bR24) and to reproduce that I needed a newer version of `wp-show` that actually removes the elements from the DOM. So I added it, but then the tests of `wp-show` where not passing because the new `createTreeWalker` algorithm doesn't support templates, so I started adding support for that. That part of the code is still a mess and I'm not even sure if it works, so I'll keep working on that on Monday, or I'll remove the code and I'll skip the test until we add this in another PR.

The good news is that the new `useSignalEffect` passes the tests and the older one doesn't so it's an improvement 🙂

luisherranz (approved):

luisherranz:
I couldn't use `data-wp-bind.hidden` because then the test that checks that effects are removed when the DOM element is removed fails, but it turns out that the tests pass with the current implementation of `wp-show` as well.

This should be ready now. As you started the PR, I've preapproved it. Feel free to review and merge 🙂

luisherranz:
Oh, the tests pass on my computer but not here. I don't know what's going on.

luisherranz:
Well, I ended up:

- Adding a temporary fake `wp-show` directive that removes things from the DOM because the other was failing in webkit.
- Testing if something was added to the DOM with state and an element instead of checking a value in `window` because it was a bit flaky.

Things should be ready now.

src/runtime/directives.js
```diff
@@ -58,7 +58,7 @@ export default () => {
 		const contextValue = useContext(context);
 		Object.entries(on).forEach(([name, path]) => {
 			element.props[`on${name}`] = (event) => {
-				evaluate(path, { event, context: contextValue });
+				return evaluate(path, { event, context: contextValue });
```
DAreRodz: This return is not needed; I'm pretty sure event listener callbacks always return `undefined` (see [this](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#the_event_listener_callback:~:text=has%20occurred%2C%20and-,it%20returns%20nothing,-.)). I'll change this, don't worry. 🙂

DAreRodz (commented):
Thanks @luisherranz! The changes you did look great to me. 😄

src/runtime/directives.js
luisherranz: Oh, I see. So instead of returning `false` in the `onclick` attribute, you'd use `event.stopPropagation()` in the event listener. And Preact is using the listener, not the attribute, right?

PR #227, opened 2023-04-24
SSR: Don't process directives in inner blocks

ockham:
_Extracted from #215. See https://github.com/WordPress/block-interactivity-experiments/pull/215#issuecomment-1519743685 for the rationale._

Since we are [hooking `wp_process_directives` into `render_block`](https://github.com/WordPress/block-interactivity-experiments/blob/00f7825934ec58ed6482c2a7548f73a27d65f838/wp-directives.php#L251-L277), it gets run repeatedly on markup inside nested inner blocks (once for each level of block nesting around a given piece of markup).

### DIFF
--- a/wp-directives.php
+++ b/wp-directives.php
@@ -233,28 +233,39 @@ add_filter( 'render_block', 'wp_directives_mark_interactive_blocks', 10, 3 );
  */
 function wp_directives_inner_blocks( $parsed_block, $source_block, $parent_block ) {
 	if (
-		isset( $parent_block ) &&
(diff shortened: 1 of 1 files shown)


ockham:
On second glance, I've noticed that with this patch applied on top of #141 and tested with the Movies Demo, we're actually _missing_ one `<template>` wrapper that the `wp-show` directive processor is supposed to wrap around the `<div>` with `data-wp-show="selectors.wpmovies.isVideosTab"` 🤦‍♂️ 

Closing.

PR #232, opened 2023-05-17
Support negation operator in selectors

SantosGuillamot:
Right now, it seems that the negation operator when using selectors doesn't work as expected. So using `data-wp-bind.hidden="!selectors.show"` doesn't work. If I am not mistaken, it is because we are not considering the option that it can be a function. Because of that, I moved the `hasNegationOperator` logic to the `getEvaluate` function.

Apart from that, I included a couple of basic tests to ensure it works in both cases.

Let me know if this is the correct approach or if we should do something different.

### DIFF
--- a/e2e/html/negation-operator.html
+++ b/e2e/html/negation-operator.html
@@ -1,0 +1,29 @@
+<!DOCTYPE html>
+<html>
+	<head>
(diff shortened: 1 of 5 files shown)


luisherranz (approved):
Looks perfect to me! 🎉

PR #233, opened 2023-05-19
Support enumerated attributes that can contain true or false

SantosGuillamot:
In the `data-wp-bind` directive we are assuming that, when the value returns `true/false` it should be a [boolean attribute](https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML). However, there are some [enumerated attributes](https://developer.mozilla.org/en-US/docs/Glossary/Enumerated) that can contain true or false as well (mainly the[ aria attributes](https://developer.mozilla.org/en-US/docs/Glossary/Boolean/ARIA)):

* Boolean attributes: It is true when it is present, and false when it is absent.
* Enumerated attributes: Booleans only come in the form of strings "true" and "false".

For example, if we compare the `hidden` attribute (boolean), with the `aria-hidden` attribute (enumerated):

```html
<div hidden aria-hidden="true">Hidden element</div>
<div aria-hidden="false">Not hidden element</div>
```

At this moment, we are handling everything as boolean attributes, and this pull request aims to fix that.

### DIFF
--- a/e2e/html/directive-bind.html
+++ b/e2e/html/directive-bind.html
@@ -45,6 +45,16 @@
 			Update
 		</button>
 
+		<p
+			data-wp-bind.hidden="!state.show"
+			data-wp-bind.aria-hidden="!state.show"
+			data-wp-bind.aria-expanded="state.show"
+			data-wp-bind.draggable="state.show"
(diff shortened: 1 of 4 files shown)


luisherranz (approved):
Perfect 👏 🙂

PR #235, opened 2023-05-19
Add priorities to directives

DAreRodz:
## What

This PR extends the `directive()` function API to add a new option called `priority`, making directives **on the same element** to be evaluated in priority order. In such an element, directives with the same priority are evaluated together, wrapped in the same component, and those with less priority are wrapped in a nested component.

Priorities follow the same schema as in WordPress, i.e., priority is a numeric value, lower numbers correspond with earlier execution ("more" priority), and the default value is `10`.

Note that this value is per directive type, as it is set during directive definition. It doesn't allow changing the order in which directives are evaluated on different parts of the HTML.

A trivial example:

```js
directive(
  'customdirective',
  ({ directives: { customdirective }) => { /* ... */ },
  { priority: 12 } // arbitrary value, just for display purposes
);
```

In addition, the priority for the `data-wp-context` directive has been set to `5`.

## Why

Priorities allow directives to manipulate the element in which they appear, in the correct order. It is especially convenient for directives that add wrappers, like a context `Provider` (e.g. `data-wp-context`), allowing other directives with less priority in the same element to consume that context.

## How

The main changes are in `/src/runtime/hooks.js`:
- I've implemented a hook called `usePriorityLevels`, that receives the object with all directives defined in an element and returns them sorted and grouped by priority.
- I'm using a recursive component called `RecursivePriorityLevel` inside the `Directive` component. It renders a component instance for each priority level that `usePriorityLevels` yields.

### DIFF
--- a/e2e/js/directive-priorities.js
+++ b/e2e/js/directive-priorities.js
@@ -1,0 +1,104 @@
+import { useEffect, useContext } from 'preact/hooks';
+import { deepSignal } from 'deepsignal';
+
+import { store } from '../../src/runtime/store';
+import { directive } from '../../src/runtime/hooks';
+
+/**
+ * Util to check that render calls happen in order.
+ */
+const executionProof = (n) => {
+	const el = document.querySelector('[data-testid="execution order"]');
+	if (!el.textContent) el.textContent = n;
+	else el.textContent += `, ${n}`;
+};
+
+/**
+ * Simple context directive, just for testing purposes. It provides a deep
+ * signal with these two properties:
+ * - attribute: 'from context'
+ * - text: 'from context'
+ */
+directive(
+	'test-context',
+	({ context: { Provider }, props: { children } }) => {
+		executionProof('context');
+		const value = deepSignal({
+			attribute: 'from context',
+			text: 'from context',
+		});
+		return <Provider value={value}>{children}</Provider>;
+	},
+	{ priority: 8 }
+);
+
+/**
+ * Simple attribute directive, for testing purposes. It reads the value of
+ * `attribute` from context and populates `data-attribute` with it.
+ */
+directive('test-attribute', ({ context, evaluate, element }) => {
+	executionProof('attribute');
+	const contextValue = useContext(context);
+	const attributeValue = evaluate('context.attribute', {
+		context: contextValue,
+	});
+	useEffect(() => {
+		element.ref.current.setAttribute('data-attribute', attributeValue);
+	}, []);
+	element.props['data-attribute'] = attributeValue;
+});
+
+/**
+ * Simple text directive, for testing purposes. It reads the value of
+ * `text` from context and populates `children` with it.
+ */
+directive(
+	'test-text',
+	({ context, evaluate, element }) => {
+		executionProof('text');
+		const contextValue = useContext(context);
+		const textValue = evaluate('context.text', {
+			context: contextValue,
+		});
+		element.props.children = <p data-testid="text">{textValue}</p>;
+	},
+	{ priority: 12 }
+);
+
+/**
+ * Children directive, for testing purposes. It adds a wrapper around
+ * `children`, including two buttons to modify `text` and `attribute values from
+ * the received context.
+ */
+directive(
+	'test-children',
+	({ context, evaluate, element }) => {
+		executionProof('children');
+		const contextValue = useContext(context);
+		const updateAttribute = () => {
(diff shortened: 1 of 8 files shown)


SantosGuillamot (approved):
Approving from my side because I tested and it seems to work great 🙂  (actually, it can help fixing [this issue](https://github.com/WordPress/gutenberg/issues/50744)).

I just tested that it works. Please wait for a proper code review before merging it.

src/runtime/hooks.js
```diff
@@ -34,12 +36,79 @@ const getEvaluate =
...
+const usePriorityLevels = (directives) =>
+	useMemo(() => {
+		const byPriority = Object.entries(directives).reduce(
+			(acc, [name, values]) => {
+				const priority = directivePriorities[name];
+				if (!acc[priority]) {
+					acc[priority] = { [name]: values };
+				} else {
+					acc[priority][name] = values;
+				}
```
luisherranz:
```suggestion
				if (!acc[priority]) acc[priority] = {};
				acc[priority][name] = values;
```

src/runtime/hooks.js
```diff
@@ -34,12 +36,79 @@ const getEvaluate =
...
+const RecursivePriorityLevel = ({
+	directives: [directives, ...rest],
+	element,
+	evaluate,
+	originalProps,
+}) => {
+	// This element needs to be a fresh copy so we are not modifying an already
+	// rendered element. This prevents an error with changes in
+	// `element.props.children` not being reflected in `element.__k`.
+	element = cloneElement(element);
```
luisherranz: @DAreRodz, can you explain this in more detail? Why is `element` already rendered?

src/runtime/hooks.js
```diff
@@ -34,12 +36,79 @@ const getEvaluate =
...
+
+	// Recursively render the wrapper for the next priority level.
+	//
+	// Note that, even though we're effectively calling `createElement` here,
```
luisherranz: What do you mean by "we're effectively calling `createElement` here"?

e2e/js/directive-priorities.js
```diff
@@ -0,0 +1,104 @@
...
+		});
+		element.props.children = <p data-testid="text">{textValue}</p>;
+	},
+	{ priority: 12 }
+);
+
+/**
+ * Children directive, for testing purposes. It adds a wrapper around
+ * `children`, including two buttons to modify `text` and `attribute values from
+ * the received context.
```
luisherranz:
```suggestion
 * Children directive, for testing purposes. It adds a wrapper around
 * `children`, including two buttons to modify `text` and `attribute` values
 * from the received context.
```

e2e/specs/directive-priorities.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+			name: 'Update attribute',
+		});
+		const updateText = element.getByRole('button', {
+			name: 'Update text',
+		});
+
+		// Modify `attribute` inside context. This triggers a re-render for the
+		// component that wraps the `attribute` directive, evaluating it again.
+		// Nested components are re-rendered as well, so their directives are
+		// also re-evaluated (note how `text` and `children` have run).
```
luisherranz, line 44: Maybe we can optimize this in the future.

luisherranz (approved):
Looks good to me as well. Great work, David 🙂

src/runtime/hooks.js
DAreRodz: My bad, I meant `h()` 😅 . I wanted to say that we are creating a vnode with a `RecursivePriorityLevel` element. I have updated the comment to make it clearer.

src/runtime/hooks.js
DAreRodz: The thing is that `element` comes from the `Directive` component, which never re-renders, so `element` is always the same instance. If we don't clone it, it's unsafe to modify its properties the way we do inside directives because some element's internals wouldn't be updated (like the `__k` property).

DAreRodz:
Thanks for the reviews. PR is merged now. 🙌 

As a follow-up, we can try improving the code with these changes:
- Move the logic inside the `usePriorityLevels` hook to the `vnode` option hook.
- Refactor components so we don't need an extra component (`Directive`)
- Show directive names for each `RecursivePriorityLevel` in the preact dev tools

src/runtime/hooks.js
luisherranz: Ok, thanks!

PR #236, opened 2023-05-22
Follow preact logic on enumerated attributes

luisherranz:
## What

I have replaced the logic of the enumerated attributes to match Preact's logic.

## Why

I was having second thoughts about including `contenteditable`, `draggable` and `spellcheck` in the codebase. If we do that, we'd need to add all the exceptions, when those cases can be handled by returning the `"true"` and `"false"` strings.

Also, whatever we do needs to match what Preact does to avoid inconsistencies.

## How

[This is Preact's logic](https://github.com/preactjs/preact/blob/ea49f7a0f9d1ff2c98c0bdd66aa0cbc583055246/src/diff/props.js#L131C24-L136), so just replace our logic with that one that only treats `aria-` and `data-` as special enumerated attributes.

### DIFF
--- a/e2e/html/directive-bind.html
+++ b/e2e/html/directive-bind.html
@@ -49,7 +49,7 @@
 			data-wp-bind.hidden="!state.show"
 			data-wp-bind.aria-hidden="!state.show"
 			data-wp-bind.aria-expanded="state.show"
-			data-wp-bind.draggable="state.show"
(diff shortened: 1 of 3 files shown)


SantosGuillamot (approved):
Looks great! 🙂 It totally makes sense to me to follow Preact approach in this.

PR #239, opened 2023-05-31
Rename sufix character and island directive

### LINKED ISSUE #132: Should we follow the HTML specification to the letter?
Hi everyone,

I am wondering if you have considered the fact, that the directives introduced here are not valid html when they are used as an attribute?

On regular, built in html elements you are not allowed to define arbitrary attributes. In these cases, you have to use a `data-` attribute to keep the markup valid.

For custom elements you are basically free to use whatever you want.

Although most JS frameworks use custom attributes when coding, will compile them away / remove these attributes when rendering to the DOM or on the server.

If you are aware of this issue, was it a conscious choice to still use those directives as attributes? If so, what was the reasoning? While there is an ongoing discussion about how important valid markup is, I think this will be a sensible topic in the WordPress ecosystem.

If you do not intend to use the directives as attributes, I apologise for raising this issue. It is still not so easy to get all of the details right for this project from the outside.

### LINKED ISSUE #238: Rename `data-wp-island` to `data-wp-interactive`
We inherited the name from the JavaScript ecosystem, but it's true that there's no context in WordPress to understand what is an island. So maybe we should rename it to `data-wp-interactive`. What do you think?

_Changed suggested by @carolinan [here](https://github.com/WordPress/gutenberg/pull/50373#pullrequestreview-1443037163)._

luisherranz:
## What

1. I've renamed the directive syntax from `data-wp-on.click` to `data-wp-on--click`.

   - Fixes https://github.com/WordPress/block-interactivity-experiments/issues/132.

2. I've renamed the `data-wp-island` directive to `data-wp-interactive`.

   - Fixes https://github.com/WordPress/block-interactivity-experiments/issues/238

## Why

1. Because JSX doesn't support attributes with dots in them.
2. Because the term _island_ has no meaning in WordPress, and we don't intend to embrace it.

## How

1. I just renamed everything, including the regular expression.

   I also added an explanation of the expression and made sure that it remains valid HTML by disallowing anything other than alphanumeric characters and hyphens. I purposely disallowed underscores to prevent confusion with hyphens.

   I also renamed the prefix constant to hardcode `data-` and ensure that it always generates valid HTML.

2. I just renamed `data-wp-island` to `data-wp-interactive`.

### DIFF
--- a/e2e/html/csn-page-2.html
+++ b/e2e/html/csn-page-2.html
@@ -14,7 +14,7 @@
 		<h1 data-testid="csn-heading-page-2">Client-side navigation Page 2</h1>
 		<h2 data-testid="csn-subheading">Subheading</h2>
 		<button
-			data-wp-on.click="actions.toggleNewValue"
+			data-wp-on--click="actions.toggleNewValue"
 			data-testid="toggle newValue"
 		>
 			Toggle trueValue
@@ -29,7 +29,7 @@
 		</div>
 
 		<button
-			data-wp-on.click="actions.replaceWithPage3"
+			data-wp-on--click="actions.replaceWithPage3"
 			data-testid="replace with page 3"
 		>
 			Replace with page 3
--- a/e2e/html/directive-bind.html
+++ b/e2e/html/directive-bind.html
@@ -6,50 +6,50 @@
 	</head>
 	<body>
 		<a
-			data-wp-bind.href="state.url"
+			data-wp-bind--href="state.url"
 			data-testid="add missing href at hydration"
 		></a>
 
 		<a
 			href="/other-url"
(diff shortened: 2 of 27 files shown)


SantosGuillamot (approved):
I've tested it, and it seems to be working fine 👌 Thanks for taking care of it!

Just a minor detail: should we change the comments in the directives definition ([link](https://github.com/WordPress/block-interactivity-experiments/blob/aa504202ccbc717045c95fa3ce4cc4feadff40c2/src/runtime/directives.js#L57))?

luisherranz:
> should we change the comments in the directives definition (link)?

Oh, true. I'll do that, thanks 🙂

PR #241, opened 2023-06-14
Experiment: Add view transitions API directive

SantosGuillamot:
I started exploring the possibility of creating a directive for the View Transitions API. Here you have a video with the explanation:

https://www.loom.com/share/39631b65aebf42b386f0fee14e0b3f58?sid=fd5f26d5-1105-43ab-87a0-d66467e6ebff

This is the [branch](https://github.com/WordPress/wp-movies-demo/tree/view-transitions-support) of the movies repo.

Basically, what I did:

* Support a new property `viewTransitionsAPI` in the `wp-link` directive. If it is set to true, the viewTransitionsAPI is enabled while navigating. This already enables all the transitions that can be achieved only with CSS.
* For more complex uses cases, I created a `data-wp-transition` directive:
    * It creates a global action and global state to be able to create transitions that need a unique id. The transitions are added dynamically with JS. When the action is called, it updates the global state and add the transition to the current element. In the new page, we have a `init` function that checks if the key defined in the directive matches the global state. If so, it adds the transition dynamically.

There are some aspects to review, although it is already working more or less fine for the movies:
* There seems to be a problem with the `backdrop-filter`.
* Adding the global action to the same element using the `wp-link` directive seems to cause a conflict.
* We have to rethink the naming and structure of the new directives.

### DIFF
--- a/src/runtime/directives.js
+++ b/src/runtime/directives.js
@@ -80,10 +80,6 @@ export default () => {
 							navigation: {
 								addTransition: ({ ref, state }) => {
 									// Reset the previous element transition
-									console.log(
-										state.core.navigation
-											.viewTransitionElement
-									);
 									if (
 										state.core.navigation
 											.viewTransitionElement
@@ -92,23 +88,8 @@ export default () => {
 											'';
 
 									// Get new key
-									let linkKey = '';
-									if (
-										ref.getAttribute('data-wp-transition')
-									) {
-										linkKey =
-											ref.getAttribute(
-												'data-wp-transition'
-											);
-									} else if (
-										ref.getAttribute(
-											'[data-wp-transition--main]'
-										)
-									) {
-										linkKey = ref.getAttribute(
-											'data-wp-transition--main'
-										);
-									}
+									const linkKey =
+										ref.getAttribute('data-wp-transition');
 
 									ref.style.viewTransitionName = linkKey;
 									state.core.navigation.viewTransitionKey =
@@ -122,12 +103,12 @@ export default () => {
 				});
 
 			const contextValue = useContext(context);
-			Object.entries(transition).forEach(([option, key]) => {
+			Object.values(transition).forEach((key) => {
 				useLayoutEffect(() => {
 					// If there is no Key, set it to the main option if exists.
 					if (
 						!state.core.navigation.viewTransitionKey &&
-						option === 'main'
+						element.props['data-wp-transition-main']
 					) {
 						state.core.navigation.viewTransitionKey = key;
 						state.core.navigation.viewTransitionElement =
--- a/src/runtime/directives.js
+++ b/src/runtime/directives.js
@@ -167,21 +167,10 @@ export default () => {
 						link.viewTransitionsAPI
 					) {
 						document.startViewTransition(async () => {
-							await navigate(href);
+							await navigate(href, link?.scroll);
 						});
 					} else {
-						await navigate(href);
-					}
-
-					// Update the scroll, depending on the option. True by default.
-					if (link?.scroll === 'smooth') {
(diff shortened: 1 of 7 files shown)


luisherranz:
Great exploration 🙂

It'd be cool to add this functionality ad-hoc to the movies demo so people can see what's possible, but wait more to merge it until we figure out the best way to do it for any theme.

Maybe we can add the `data-wp-view-transition` directives to the movies demo plugin and in this codebase expose `store` in the directives and keep the support for `viewTransition` in `data-wp-link`. 

I think we won't port `data-wp-link` as it is to Gutenberg because we should probably switch to an event delegation logic, but we can keep using it here.

SantosGuillamot:
> It'd be cool to add this functionality ad-hoc to the movies demo so people can see what's possible, but wait more to merge it until we figure out the best way to do it for any theme.

That makes sense to me 🙂 I'll try to work on that in the upcoming weeks.

SantosGuillamot:
I've moved the `data-wp-transition` to the movies repo PR. I had to exposed some functions in order to use them there.

luisherranz (approved):
LGTM! 🎈

luisherranz:
Let's add this change to start using `data-wp-key`: https://github.com/WordPress/wp-movies-demo/pull/52/files#r1247615718

src/runtime/directives.js
```diff
@@ -161,9 +161,17 @@ export default () => {
...
 					event.preventDefault();
-
 					// Fetch the page (or return it from cache).
-					await navigate(href);
+					if (
+						document.startViewTransition &&
+						link.viewTransitionsAPI
+					) {
+						document.startViewTransition(async () => {
+							await navigate(href);
```
luisherranz: Let's move the scroll here as well, so the transitions is done taking into account the new position.

wp-directives.php
```diff
@@ -128,10 +128,10 @@ function wp_directives_add_wp_link_attribute( $block_content ) {
 			) {
 				$w->set_attribute(
 					'data-wp-link',
-					'{ "prefetch": true, "scroll": false }'
+					'{ "prefetch": true, "viewTransitionsAPI": true, "scroll": false }'
 				);
 			} else {
-				$w->set_attribute( 'data-wp-link', '{ "prefetch": true }' );
+				$w->set_attribute( 'data-wp-link', '{ "prefetch": true, "viewTransitionsAPI": true, "scroll": false }' );
```
luisherranz: Let's recover the scroll and see if it works fine when we do it inside the `startTransition`.

src/runtime/directives.js
SantosGuillamot: I wasn't sure if it was the correct option because I don't know where navigate is used, but I moved the scroll to the navigate function. I thought that maybe external blocks using the navigate function want that option as well. And I was basically repeating the code twice inside the `startViewTransition` and outside of it.

Anyway, happy to move it if that's not the correct place.

luisherranz: Does it solve the scroll problem during transtions?

SantosGuillamot: I think so. It looks much better now. I believe this is the expected behavior:


https://github.com/WordPress/block-interactivity-experiments/assets/34552881/6f681e54-5325-407d-ab2b-d1a199c45fcc

luisherranz: Great! 🙂

luisherranz:

These changes were already deployed to the [wp-movies-demo](https://github.com/WordPress/wp-movies-demo) site, but we won't merge them until we figure out better how to integrate the new View Transitions API. Please, go to [this discussion](https://github.com/WordPress/gutenberg/discussions/53063) to continue the conversation.

